PR #10, opened 2018-10-17
callback-based retryImageAction + Don't ignore errors reported during Docker push/pull

mtrmac:
Per conversation in https://github.com/openshift/origin/pull/20941, this:

- Modifies `RetryImageAction`→`retryImageAction` to use a callback instead of containing parts of the calling functions and using reflection to choose which part to call. This could reportedly be useful for the `buildah`-based implementation as well.

- Fixes `dockerPushImage`/`dockerPullImage` to correctly detect errors reported through the `imageprogress` channel. These functions don’t have any callers outside of tests anymore, so this is not really interesting — but this preserves the bug fixes in case they were revived, and I needed to updated them for the `retryImageAction` changes to build anyway.

Manually tested only using `go test ./...`; hoping the CI can handle the bigger picture.

### DIFF
--- a/pkg/build/builder/dockerutil.go
+++ b/pkg/build/builder/dockerutil.go
@@ -9,7 +9,6 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
-	"reflect"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


bparees:
@nalind fyi this presumably will conflict w/ your effort to reintroduce the retry logic in our push/pull flows.

/lgtm

mtrmac:
@nalind If you already have everything you need and are dropping the `dockerPullImage`/`dockerPushImage` implementations, feel free to just close this; I don’t need this to be _my_ PR. Whatever is easiest for you to work with.

PR #11, opened 2018-10-17
Retry pulling or pushing images again, for certain errors

nalind:
Replace direct calls to `dockerClient.PullImage()`/`dockerClient.PushImage()` with calls to `dockerPullImage()`/`dockerPushImage()`, which wrap them using `retryImageAction()`, so that we retry them if the fail due to what appears to be a transient error.

This means that we don't get to check if the passed-in reference is a digested or tagged reference, so the daemonless `PushImage()`/`PullImage()` methods can receive a digest as a `Tag` value again.  To cope with that, check if a "tag" can be parsed as a valid digest, and if it can be, just treat it as a digest.


nalind:
Hmm, going to try something less ambitious.

PR #14, opened 2018-10-18
only enable logrus debug at higher loglevels

bparees:


bparees:
@adambkaplan ptal

pkg/build/builder/cmd/builder.go
```diff
@@ -314,49 +314,40 @@ func runBuild(out io.Writer, builder builder) error {
...
-		case glog.Is(4):
-			serviceability.InitLogrus("DEBUG")
-		case glog.Is(2):
-			serviceability.InitLogrus("INFO")
-		case glog.Is(0):
-			serviceability.InitLogrus("WARN")
-		}
-	*/
+	switch {
+	case glog.Is(6):
```
adambkaplan: Why level 6 vs. 4 for DEBUG level? Isn't `V(5)` our standard level for verbose logs?

bparees: buildah's logrus output is *very* noisy.  I want a way to enable our standard level 5 logging without turning on the super verbose buildah output.

so doing this lets us have that distinction.

PR #15, opened 2018-10-18
Copy CAs mounted via ConfigMap

adambkaplan:
JIRA IDs: DEVEXP-181, DEVEXP-182

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -14,9 +14,9 @@ import (
 
 	"k8s.io/apiserver/pkg/util/logs"
 
-	"github.com/openshift/library-go/pkg/serviceability"
-
+	"github.com/openshift/builder/pkg/build/builder"
 	"github.com/openshift/builder/pkg/version"
+	"github.com/openshift/library-go/pkg/serviceability"
 )
 
 func main() {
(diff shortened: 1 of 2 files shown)


pkg/build/builder/util.go
```diff
@@ -22,6 +22,12 @@ import (
 	builderutil "github.com/openshift/builder/pkg/build/builder/util"
 )
 
+// Mount paths for certificate authorities
+const (
+	ConfigMapCertsMountPath = "/var/run/configs/openshift.io/certs"
```
adambkaplan: ConfigMap with CAs will mount here

cmd/main.go
```diff
@@ -34,33 +34,55 @@ func main() {
...
+	clusterCASrc := fmt.Sprintf("%s/ca.crt", builder.SecretCertsMountPath)
+	clusterCADst := fmt.Sprintf("%s/cluster.crt", tlsCertRoot)
+	err := CopyIfExists(clusterCASrc, clusterCADst)
+	if err != nil {
+		fmt.Printf("Error setting up cluster CA cert: %v", err)
+		os.Exit(1)
+	}
+
+	// TODO: Remove this once the config-map based mount approach lands after rebase
+	oldServiceCASrc := fmt.Sprintf("%s/service-ca.crt", builder.SecretCertsMountPath)
```
adambkaplan: This should support both mount options for the service signing CA concurrently

PR #16, opened 2018-10-18
add the new registry service to the insecure list

bparees:


bparees:
@adambkaplan this is the change you'll need to be able to push/pull from the internal registry, if indeed service-ca.crt is no longer being automatically mounted into build pods.

bparees:
@nalind we've tentatively lined up @adambkaplan to demo in your place in case you run into issues, but fyi you'll need this change if you intend to run a build on the cluster that either pulls or pushes to the internal registry.

(you can pull/push to docker.io of course, but then you need to supply a push/pull secret).

Also your build cannot reference images via imagestream(for reasons i can explain elsewhere), so your build object needs to use DockerImage references like:

```
  output:
    to:
      kind: DockerImage
      name: image-registry.openshift-image-registry.svc:5000/bparees/nodejs-ex
```

and:

```
  strategy:
    sourceStrategy:
      from:
        kind: DockerImage
        name: docker.io/centos/nodejs-8-centos7:latest
```

bparees:
(the reason this change is needed is because the service-ca.crt is no longer being made available inside build pods, so the logic i added here: https://github.com/openshift/builder/blob/master/cmd/main.go#L46-L53 is no longer effective)

nalind:
@bparees I'm skipping TLS setup for the moment - `oc adm registry` got the registry running after `hack/start-local.sh` brought it up, the node's dockerd had been configured to not check certificates on the registry, and I created a `build-system-configs` configmap that marked the internal registry the same way for builds.

After that, building, pushing, and deploying the nodejs-ex and simple-openshift-sinatra-sti samples succeeded for me (per https://gist.github.com/nalind/c2b906126ef833aeb030837856d7ea5c).

Will definitely call out that doing it right (with TLS properly configured) is something we're still working on.

bparees:
> @bparees I'm skipping TLS setup for the moment - oc adm registry got the registry running after hack/start-local.sh brought it up, 

unfortunately the demos are supposed to be done using the official openshift installer:
https://github.com/openshift/installer
https://github.com/openshift/installer/blob/master/docs/dev/libvirt-howto.md

(at which point you'll run into the registry issues I emailed about, as well as the insecure issue i'm "fixing" with this PR)

> the node's dockerd had been configured to not check certificates on the registry, and I created a build-system-configs configmap that marked the internal registry the same way for builds.

I don't understand this part...  if you did install the registry using oc adm registry, then your registry should be "docker-registry.default.svc" in which case the builder image already treats that registry as insecure.  (unless you're accessing the registry by ip address? In which case i guess you added the registry's ip address in your registries.conf that you put in your configmap?)

nalind:
Yes, it's listing the registry by IP address.  The installer hasn't been giving me good results with libvirt.

bparees:
@nalind which days are you working between now and wednesday?

nalind:
@bparees I'm OOTO for all of them.

bparees:
> @bparees I'm OOTO for all of them.

ok, let's just plan to have @adambkaplan demo in your place, he should be able to do it on a real cluster and, if this PR merges, with less "hacking" to get the trust in place.

bparees:
(unless you really want to, i don't want to take the opportunity away from you and i know you've already put the effort into getting a demo together)

nalind:
I'll take you and @adambkaplan up on that, much appreciated!

PR #20, opened 2018-10-25
remove vestiges of docker daemon codepaths

bparees:
this also sets the output image digest again so we can re-enable some more tests.


gabemontero:
fyi viewing the file deltas with `?w=1` makes a big difference :-)

generally looks good but I'm not sure if my bystander level focus to date on openshift/builder warrants me placing the lgtm 

@openshift/sig-developer-experience

bparees:
/assign @adambkaplan 
@nalind fyi

pkg/build/builder/util.go
```diff
@@ -85,90 +80,6 @@ func addBuildLabels(labels map[string]string, build *buildapiv1.Build) {
...
-	} else if err != nil {
-		return -1, err
-	}
-	return val, nil
-}
-
-// readNetClsCGroup parses /proc/self/cgroup in order to determine the container id that can be used
-// the network namespace that this process is running on, it returns the cgroup and container type
-// (docker vs crio).
-func readNetClsCGroup(reader io.Reader) (string, string) {
```
adambkaplan: @bparees these are now the concern of buildah (or completely irrelevant)?

bparees: irrelevant.  Since the container is running inside the pod, it inherits the network cgroup as i understand it.  We needed this previously because the container ran outside the pod, so it picked up the host's network.

glide.yaml
```diff
@@ -13,7 +13,7 @@ import:
   - builder/dockerfile/parser
 - package: github.com/fsouza/go-dockerclient
   repo:    https://github.com/openshift/go-dockerclient.git
-  version: openshift-3.9
+  version: openshift-4.0
```
adambkaplan: reason for not floating on `master`?

bparees: in general we never want to float on master if we have a choice.  Even the other ones i switched to master is because they don't have release branches and at some point (closer to shipping) we'll probably want to switch them to specific commits so they don't get updated when someone updates a dep.

pkg/build/builder/cmd/builder.go
```diff
@@ -92,60 +92,51 @@ func newBuilderConfigFromEnvironment(out io.Writer, needsDocker bool) (*builderC
...
-				return nil, fmt.Errorf("no daemonless store: %v", err)
-			}
-			cfg.dockerClient = dockerClient
-			cfg.dockerEndpoint = "n/a"
+		dockerClient, err := bld.GetDaemonlessClient(systemContext, store, os.Getenv("BUILD_ISOLATION"))
+		if err != nil {
+			return nil, fmt.Errorf("no daemonless store: %v", err)
 		}
+		cfg.dockerClient = dockerClient
+		cfg.dockerEndpoint = "n/a"
```
adambkaplan: Any reason why this can't be the empty string?

bparees: that's a question for @nalind but this PR retained the existing value.

nalind: IIRC we used to hit an error in the S2I Validator() function (somewhere around https://github.com/openshift/source-to-image/blob/master/pkg/api/validation/validation.go#L23) if we just left it unset.

adambkaplan: @bparees probably worth a comment, then (and a future techdebt fix for s2i).

bparees: commented.

PR #21, opened 2018-11-02
Renaming additional ca file

adambkaplan:
* Aligns with openshift/origin#21362


cmd/main.go
```diff
@@ -61,8 +61,8 @@ func main() {
 		os.Exit(1)
 	}
 
-	additionalCASrc := fmt.Sprintf("%s/trusted-ca.crt", builder.ConfigMapCertsMountPath)
-	additionalCADst := fmt.Sprintf("%s/openshift-trusted-ca.crt", tlsCertRoot)
+	additionalCASrc := fmt.Sprintf("%s/additional-ca.crt", builder.ConfigMapCertsMountPath)
```
adambkaplan: @deads2k are we better off putting this constant in the API?

bparees: it's not a user-facing api, it's an internal implementation detail of the controller/builder contract, so i would say no, it does not belong in the api.

cmd/main.go
```diff
@@ -61,8 +61,8 @@ func main() {
...
+	additionalCADst := fmt.Sprintf("%s/openshift-additional-ca.crt", tlsCertRoot)
```
bparees: why are we changing the name between the source and destination?

adambkaplan: I kept the pattern with the service signing cert, though this probably is not necessary.

cmd/main.go
adambkaplan: Don't see too many options other than policing ourselves/e2e testing - this is too simple to be added to`library-go`, and we don't have other shared repos that fit this use case.

deads2k: > Don't see too many options other than policing ourselves/e2e testing - this is too simple to be added to`library-go`, and we don't have other shared repos that fit this use case.

We don't vendor for constants.

cmd/main.go
```diff
@@ -61,8 +61,8 @@ func main() {
 		os.Exit(1)
 	}
 
-	additionalCASrc := fmt.Sprintf("%s/trusted-ca.crt", builder.ConfigMapCertsMountPath)
-	additionalCADst := fmt.Sprintf("%s/openshift-trusted-ca.crt", tlsCertRoot)
+	additionalCASrc := fmt.Sprintf("%s/additional-ca.crt", builder.ConfigMapCertsMountPath)
+	additionalCADst := fmt.Sprintf("%s/additional-ca.crt", tlsCertRoot)
```
adambkaplan: @bparees fixed file name.

PR #27, opened 2018-11-28
add retry logic for s2i pushes

bparees:


bparees:
@smarterclayton fyi.  we had this for docker strategy builds but it was missing on s2i builds.

bparees:
failing test not related.

PR #30, opened 2018-12-07
Revert "copy build certificates to /etc/docker/certs.d"

bparees:
Reverts openshift/builder#29

PR #31, opened 2018-12-07
Restore "copy build certificates to /etc/docker/certs.d"

adambkaplan:
Restores functionality in #29
Reverts #30

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -3,6 +3,7 @@ package main
 import (
 	"fmt"
 	"io"
+	"io/ioutil"
 	"math/rand"
 	"os"
 	"path/filepath"
@@ -35,10 +36,11 @@ func main() {
 	}
 
 	const tlsCertRoot = "/etc/pki/tls/certs"
+	const runtimeCertRoot = "/etc/docker/certs.d"
 
 	clusterCASrc := fmt.Sprintf("%s/ca.crt", builder.SecretCertsMountPath)
 	clusterCADst := fmt.Sprintf("%s/cluster.crt", tlsCertRoot)
-	err := CopyIfExists(clusterCASrc, clusterCADst)
+	err := CopyFileIfExists(clusterCASrc, clusterCADst)
 	if err != nil {
 		fmt.Printf("Error setting up cluster CA cert: %v", err)
 		os.Exit(1)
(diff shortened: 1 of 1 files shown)


bparees:
/hold


Ben Parees | OpenShift

cmd/main.go
```diff
@@ -35,10 +36,11 @@ func main() {
 	}
 
 	const tlsCertRoot = "/etc/pki/tls/certs"
+	const runtimeCertRoot = "/etc/docker/certs.d"
 
 	clusterCASrc := fmt.Sprintf("%s/ca.crt", builder.SecretCertsMountPath)
 	clusterCADst := fmt.Sprintf("%s/cluster.crt", tlsCertRoot)
-	err := CopyIfExists(clusterCASrc, clusterCADst)
+	err := CopyFileIfExists(clusterCASrc, clusterCADst)
```
bparees: looking at a cluster today, this location exists:

ls -l /var/run/secrets/kubernetes.io/serviceaccount/ca.crt 
lrwxrwxrwx. 1 root root 13 Dec 12 20:58 /var/run/secrets/kubernetes.io/serviceaccount/ca.crt -> ..data/ca.crt

cmd/main.go
```diff
@@ -47,25 +49,26 @@ func main() {
 	// TODO: Remove this once the config-map based mount approach lands after rebase
 	oldServiceCASrc := fmt.Sprintf("%s/service-ca.crt", builder.SecretCertsMountPath)
 	oldServiceCADst := fmt.Sprintf("%s/service.crt", tlsCertRoot)
-	err = CopyIfExists(oldServiceCASrc, oldServiceCADst)
+	err = CopyFileIfExists(oldServiceCASrc, oldServiceCADst)
```
bparees: this location does not exist:
 ls -l /var/run/secrets/kubernetes.io/serviceaccount        
total 0
lrwxrwxrwx. 1 root root 13 Dec 12 20:58 ca.crt -> ..data/ca.crt
lrwxrwxrwx. 1 root root 16 Dec 12 20:58 namespace -> ..data/namespace
lrwxrwxrwx. 1 root root 12 Dec 12 20:58 token -> ..data/token

cmd/main.go
```diff
@@ -47,25 +49,26 @@ func main() {
...
 		fmt.Printf("Error setting up service CA cert: %v", err)
 		os.Exit(1)
 	}
 
+	// TODO: Remove this once the build controller mounts the internal registry's CA
+	// in certs.d/<internal-registry-hostname>/
 	newServiceCASrc := fmt.Sprintf("%s/service-ca.crt", builder.ConfigMapCertsMountPath)
 	newServiceCADst := fmt.Sprintf("%s/openshift-service.crt", tlsCertRoot)
-	err = CopyIfExists(newServiceCASrc, newServiceCADst)
+	err = CopyFileIfExists(newServiceCASrc, newServiceCADst)
```
bparees: and this, presumably, "might" exist depending on the raciness of the creation/population of the configmap that was created by the build controller.

cmd/main.go
```diff
@@ -47,25 +49,26 @@ func main() {
...
 	if err != nil {
 		fmt.Printf("Error setting up service CA cert: %v", err)
 		os.Exit(1)
 	}
 
-	additionalCASrc := fmt.Sprintf("%s/additional-ca.crt", builder.ConfigMapCertsMountPath)
-	additionalCADst := fmt.Sprintf("%s/additional-ca.crt", tlsCertRoot)
-	err = CopyIfExists(additionalCASrc, additionalCADst)
+	runtimeCASrc := fmt.Sprintf("%s/certs.d", builder.ConfigMapCertsMountPath)
+	err = CopyDirIfExists(runtimeCASrc, runtimeCertRoot)
```
bparees: similarly "might" exist depending on raciness and the level of code in the origin build controller that dictates where this gets mounted i guess.

cmd/main.go
bparees: perhaps that is not the service-ca though?  i'm not sure what it is....

cmd/main.go
bparees: so this no longer serves any purpose except confusion i guess.

bparees: (and i guess gets used on ansible installs?)

cmd/main.go
bparees: and this is only working for us because the configmap is getting populated fast enough

bparees:
passed everything that matters, manually merging.

PR #34, opened 2018-12-12
Enable blob caching and blobInfo caching

nalind:
This patch set updates the versions of the storage, image, and buildah libraries and enables blob and blobInfo caching.  We switch back to the main containers/image repository, since it now supports registry names specified as IP addresses, and drop logic for setting the InsecureSkipTLSVerify flag, which the image library's registry client logic now sets automatically if we don't.


Dockerfile
```diff
@@ -24,3 +25,6 @@ RUN ln -s /usr/bin/openshift-builder /usr/bin/openshift-sti-build && \
 LABEL io.k8s.display-name="OpenShift Builder" \
       io.k8s.description="This is a component of OpenShift and is responsible for executing image builds." \
       io.openshift.tags="openshift,builder"
+
+RUN mkdir /var/cache/blobs
+VOLUME /var/lib/containers/cache
```
adambkaplan: Is the intent that `/var/lib/containers/cache` will be mounted in by the build controller (from the host)?

pkg/build/builder/cmd/builder.go
```diff
@@ -131,7 +132,18 @@ func newBuilderConfigFromEnvironment(out io.Writer, needsDocker bool) (*builderC
 		}
 		istorage.Transport.SetStore(store)
 
-		dockerClient, err := bld.GetDaemonlessClient(systemContext, store, os.Getenv("BUILD_ISOLATION"))
+		// By default, enable a blob cache, but allow it to be moved or
+		// disabled via the environment.
+		cfg.blobCache = "/var/cache/blobs"
+		if dir, isSet := os.LookupEnv("BUILD_BLOBCACHE"); isSet {
```
adambkaplan: How will this be disabled - empty env valule?

pkg/build/builder/daemonless.go
```diff
@@ -68,28 +62,17 @@ func pullDaemonlessImage(sc types.SystemContext, store storage.Store, imageName
...
-			glog.V(2).Infof("Registry %q is marked as insecure in the registries configuration.", registry.URL)
-			systemContext.DockerInsecureSkipTLSVerify = true
-			systemContext.OCIInsecureSkipTLSVerify = true
-		} else {
-			glog.V(2).Infof("Registry %q is marked as secure in the registries configuration.", registry.URL)
-		}
-	} else {
-		glog.V(2).Infof("Registry for %q is not present in the registries configuration, assuming it is secure.", imageName)
-	}
-
```
adambkaplan: I assume the designation of secure/insecure registries will be taken care of by copying in `/etc/containers/registries.conf`?

See https://github.com/openshift/origin/pull/21653

Dockerfile
bparees: it is, but i don't think we gain anything by declaring this volume, i'd rather leave it out.  We mount stuff in lots of places in the image that aren't explicitly declared as volumes.

@nalind was there a specific reason you needed to add the volume declaration?

pkg/build/builder/cmd/builder.go
bparees: please call this BUILD_ENABLE_BLOBCACHE or something that makes it clear it's an enablement (and then make it enable when the value is set to `true` (case insensitive))

pkg/build/builder/cmd/builder.go
```diff
@@ -131,7 +132,18 @@ func newBuilderConfigFromEnvironment(out io.Writer, needsDocker bool) (*builderC
...
+			cfg.blobCache = dir
+		}
+
+		// An empty value implies the system default location, somewhere under
+		// /var/lib/containers/cache.
+		systemContext.BlobInfoCacheDir = os.Getenv("BUILD_BLOBINFOCACHE")
```
bparees: and name this one BUILD_BLOBINFOCACHE_DIR or something that makes it clear it points to a path.

bparees: also presumably you need to handle the case that someone enabled the blob cache but did not set this dir.

or do away w/ this env entirely and just hardcode the expected cache directory.

pkg/build/builder/daemonless.go
bparees: @nalind did something get fixed in buildah that we don't need this any more?

Dockerfile
nalind: It's a hint that there's a benefit to mounting `/var/lib/containers/cache` as a shared volume, but you're right that we don't do that for the other things that we mount in, so I'll drop the line.

pkg/build/builder/cmd/builder.go
nalind: Okay, changing it.

pkg/build/builder/daemonless.go
nalind: Yes, https://github.com/containers/image/pull/468 was merged.

nalind: @adambkaplan Yup, the way the builder expects the configuration to be provided to it hasn't changed, we just don't need to do this bit of work ourselves any more.

pkg/build/builder/cmd/builder.go
bparees: sorry I screwed this up.  but i think BUILD_BLOBCACHE_DIR would be a better name.

bparees: (and ignore my comments about expecting a value of `true`)

pkg/build/builder/cmd/builder.go
nalind: Okay, dropping the variable, which should result in containers/image always using its default location.

nalind:
Backing out the changes to pay attention to the ImageOptimizationPolicy, which can be done in a different PR after this one.

runcom:
This still needs a corresponding origin PR for the controller (e.g. https://github.com/openshift/origin/pull/21613)

runcom:
picked up this PR to bring in https://github.com/containers/buildah/pull/1244 for testing till it gets merged and I'm seeing errors coming from pkg/errors and raven-go. I've been able to overcome that by sticking the lib in glide with:
```
+- package: github.com/pkg/errors
+  version: 645ef00459ed84a119197bfb8d8205042c6df63d
+- package: github.com/getsentry/raven-go
+  version: 32a13797442ccb601b11761d74232773c1402d14
```
it is related to https://github.com/getsentry/raven-go/issues/225

runcom:
The parallel pull + pgzip PR in Buildah has been merged. We could grab it here, I'm already testing with it.

nalind:
> We need [containers/buildah#1265](https://github.com/containers/buildah/pull/1265) as well

Included as of https://github.com/openshift/builder/pull/34#event-2060437413.

Dockerfile
```diff
@@ -16,6 +16,8 @@ COPY --from=builder /go/src/github.com/openshift/builder/openshift-builder /usr/
 COPY imagecontent/policy.json /etc/containers/
 COPY imagecontent/registries.conf /etc/containers/
 COPY imagecontent/storage.conf /etc/containers/
+RUN mkdir /var/cache/blobs
```
bparees: There's some discussion about changing this name to "var/cache/blobs/metadata" assuming this is the metadata cache.  Please sync with @runcom and @mrunalp

bparees:
one nit on the dir name, otherwise lgtm pending passing tests.

Dockerfile
runcom: related to https://github.com/openshift/origin/pull/21613#discussion_r245744542

runcom: `/var/cache/blobs/meta` I've used that in 21613 suggested by @mrunalp

pkg/build/builder/cmd/builder.go
```diff
@@ -131,7 +132,15 @@ func newBuilderConfigFromEnvironment(out io.Writer, needsDocker bool) (*builderC
 		}
 		istorage.Transport.SetStore(store)
 
-		dockerClient, err := bld.GetDaemonlessClient(systemContext, store, os.Getenv("BUILD_ISOLATION"))
+		// Default to using /var/cache/blobs as a blob cache, but allow its location
+		// to be changed by setting $BUILD_BLOBCACHE_DIR.  Setting the location to an
+		// empty value disables the cache.
+		cfg.blobCache = "/var/cache/blobs"
```
bparees: this default should also change of course

pkg/build/builder/cmd/builder.go
```diff
@@ -131,7 +132,15 @@ func newBuilderConfigFromEnvironment(out io.Writer, needsDocker bool) (*builderC
...
+		if blobCacheDir, isSet := os.LookupEnv("BUILD_BLOBCACHE_DIR"); isSet {
```
bparees: and this env var name should change to indicate it's metadata

Dockerfile
bparees: nm, this dir is apparently for something else.

PR #35, opened 2019-01-08
Heed the ImageOptimizationPolicy

nalind:
Pass the build's configured `ImageOptimizationPolicy` down to the build logic by keeping track of it in the `DaemonlessClient`.

### DIFF
--- a/pkg/build/builder/cmd/builder.go
+++ b/pkg/build/builder/cmd/builder.go
@@ -141,7 +141,19 @@ func newBuilderConfigFromEnvironment(out io.Writer, needsDocker bool) (*builderC
 			cfg.blobCache = blobCacheDir
 		}
 
-		dockerClient, err := bld.GetDaemonlessClient(systemContext, store, os.Getenv("BUILD_ISOLATION"), cfg.blobCache)
+		imageOptimizationPolicy := buildapiv1.ImageOptimizationNone
+		if s := cfg.build.Spec.Strategy.DockerStrategy; s != nil {
+			// Default to possibly-multiple-layer builds for Dockerfile-based builds, unless something else was specified.
+			if policy := s.ImageOptimizationPolicy; policy != nil {
+				imageOptimizationPolicy = *policy
+			}
+		}
+		if s := cfg.build.Spec.Strategy.SourceStrategy; s != nil {
+			// Always use base-image+single-layer builds for S2I builds.
+			imageOptimizationPolicy = buildapiv1.ImageOptimizationSkipLayers
+		}
+
+		dockerClient, err := bld.GetDaemonlessClient(systemContext, store, os.Getenv("BUILD_ISOLATION"), cfg.blobCache, imageOptimizationPolicy)
 		if err != nil {
 			return nil, fmt.Errorf("no daemonless store: %v", err)
 		}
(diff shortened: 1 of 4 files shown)


mrunalp:
@bparees ptal

pkg/build/builder/cmd/builder.go
```diff
@@ -140,7 +140,14 @@ func newBuilderConfigFromEnvironment(out io.Writer, needsDocker bool) (*builderC
 			cfg.blobCache = blobCacheDir
 		}
 
-		dockerClient, err := bld.GetDaemonlessClient(systemContext, store, os.Getenv("BUILD_ISOLATION"), cfg.blobCache)
+		imageOptimizationPolicy := buildapiv1.ImageOptimizationNone
+		if s := cfg.build.Spec.Strategy.DockerStrategy; s != nil {
+			if policy := s.ImageOptimizationPolicy; policy != nil {
+				imageOptimizationPolicy = *policy
+			}
+		}
```
bparees: we should also set the optimization policy to "SkipLayers" when doing an s2i build.  So any time cfg.build.Spec.Strategy.SourceStrategy!=nil we should set imageOptimizationPolicy to SkipLayers.

nalind: Okay, that should be there now.  Thanks!

bparees:
/lgtm

will be exciting to see if this passes our various build tests :)

bparees:
@nalind is it possible this PR is changing the number of layers in a docker built image (even when no optimization policy is set in the buildconfig)?

you're consistently getting this failure:

```
fail [github.com/openshift/origin/test/extended/images/layers.go:203]: map[string]v1.ImageBlobReferences{"sha256:eb550fad234c7d702108ad6b72a09857ed54be298bc608c698e4703325b945ec":v1.ImageBlobReferences{ImageMissing:false, Layers:[]string{"sha256:57c14dd66db0390dbf6da578421c077f6de8e88edd0815b4caa94607ba5f4c09", "sha256:5c46b5970c3a81e83b33b9bac88e2a74bec00aad26387f6b0afa33b38442aede", "sha256:1dbcab28ce46b65c0174e5e82658492107396fead31e9144c343e6bc96e471c7", "sha256:1dbcab28ce46b65c0174e5e82658492107396fead31e9144c343e6bc96e471c7"}, Config:(*string)(0xc4214ddfc0)}}
Expected
    <int>: 4
to equal
    <int>: 2
```

That test builds a new image from a busybox (1 layer) base.

the dockerfile is:
```
FROM a
RUN mkdir -p /var/lib && echo "a" > /var/lib/file
```

yet we seem to be ending up with a 4 layer output image instead of a 2 layer one?

bparees:
(the layers test looks stable/healthy on origin, so it seems like this PR is what is disrupting it)

nalind:
Yes, that looks like a bug.  It looks like for certain directives where `docker build` creates a history entry with `empty_layer` set to true, we're going ahead and creating a layer, however small.

bparees:
/hold

this is going to keep failing until buildah stops creating extra layers when the optimization policy is set to none.

adambkaplan:
/retest

Potential infra flake (builds failed with OOMKilled)

PR #36, opened 2019-01-09
Limit Default Search to docker.io

adambkaplan:
Additional doc comments added from containers/image

### DIFF
--- a/imagecontent/registries.conf
+++ b/imagecontent/registries.conf
@@ -8,18 +8,23 @@
 # The only valid categories are: 'registries.search', 'registries.insecure',
 # and 'registries.block'.
 
+# Registries to search for images that are not fully-qualified.
+# i.e. foobar.com/my_image:latest vs my_image:latest
 [registries.search]
-registries = ['docker.io', 'registry.fedoraproject.org', 'quay.io', 'registry.access.redhat.com', 'registry.centos.org']
+registries = ['docker.io']
 
-# If you need to access insecure registries, add the registry's fully-qualified name.
-# An insecure registry is one that does not have a valid SSL certificate or only does HTTP.
+# Registries that do not use TLS when pulling images or uses self-signed
+# certificates.
 [registries.insecure]
-#registries = ['172.30.246.32:5000','172.30.1.1:5000','docker-registry.default.svc:5000','docker-registry.default.svc.cluster:5000','docker-registry.default.svc.cluster.local:5000']
-#registries = ['docker-registry.default.svc:5000','docker-registry.default.svc.cluster:5000','docker-registry.default.svc.cluster.local:5000','image-registry.openshift-image-registry.svc:5000','image-registry.openshift-image-registry.svc.cluster:5000','image-registry.openshift-image-registry.svc.cluster.local:5000']
+registries = []
 
-# If you need to block pull access from a registry, uncomment the section below
-# and add the registries fully-qualified name.
-#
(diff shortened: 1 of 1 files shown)


imagecontent/registries.conf
```diff
@@ -9,17 +9,10 @@
 # and 'registries.block'.
 
 [registries.search]
-registries = ['docker.io', 'registry.fedoraproject.org', 'quay.io', 'registry.access.redhat.com', 'registry.centos.org']
+# registries = ['docker.io', 'registry.fedoraproject.org', 'quay.io', 'registry.access.redhat.com', 'registry.centos.org']
```
bparees: i'm reasonably certain this is going to cause tests to fail unless we miraculously have fully qualified image specs for all our tests (I did recently go qualify a bunch, but for a long time we had stuff like "FROM centos").

but i guess we'll see.  Anyway you'll definitely want to investigate any aws-builds failures.

imagecontent/registries.conf
```diff
@@ -9,17 +9,10 @@
...
 [registries.insecure]
 #registries = ['172.30.246.32:5000','172.30.1.1:5000','docker-registry.default.svc:5000','docker-registry.default.svc.cluster:5000','docker-registry.default.svc.cluster.local:5000']
 #registries = ['docker-registry.default.svc:5000','docker-registry.default.svc.cluster:5000','docker-registry.default.svc.cluster.local:5000','image-registry.openshift-image-registry.svc:5000','image-registry.openshift-image-registry.svc.cluster:5000','image-registry.openshift-image-registry.svc.cluster.local:5000']
-
-# If you need to block pull access from a registry, uncomment the section below
-# and add the registries fully-qualified name.
-#
-# Docker only
-[registries.block]
-registries = []
```
bparees: i'd rather leave this (commented out) to at least provide a hint about the syntax.

imagecontent/registries.conf
```diff
@@ -9,17 +9,10 @@
...
 
 [registries.search]
-registries = ['docker.io', 'registry.fedoraproject.org', 'quay.io', 'registry.access.redhat.com', 'registry.centos.org']
+# registries = ['docker.io', 'registry.fedoraproject.org', 'quay.io', 'registry.access.redhat.com', 'registry.centos.org']
 
 # If you need to access insecure registries, add the registry's fully-qualified name.
 # An insecure registry is one that does not have a valid SSL certificate or only does HTTP.
 [registries.insecure]
 #registries = ['172.30.246.32:5000','172.30.1.1:5000','docker-registry.default.svc:5000','docker-registry.default.svc.cluster:5000','docker-registry.default.svc.cluster.local:5000']
 #registries = ['docker-registry.default.svc:5000','docker-registry.default.svc.cluster:5000','docker-registry.default.svc.cluster.local:5000','image-registry.openshift-image-registry.svc:5000','image-registry.openshift-image-registry.svc.cluster:5000','image-registry.openshift-image-registry.svc.cluster.local:5000']
```
bparees: you can update this to `registries = [ ]` too while you're at it.

adambkaplan:
@bparees updated to default search to docker.io, per conversations in api-review with @smarterclayton.

bparees:
/lgtm

keep in mind we'll still need to implement the defaulting behavior in the build controller too, so that if we overwrite this file (such as to set blocked registries) we still set the search registry to the default (if no other setting was provided)

PR #40, opened 2019-01-30
Remove docker.io as hardcoded default registry

adambkaplan:
* Per containers/image#572, buildah will still default to docker.io for image pulls.
* Prevent auth attempts if no registry is set in the image pull spec

### DIFF
--- a/pkg/build/builder/daemonless.go
+++ b/pkg/build/builder/daemonless.go
@@ -51,14 +51,21 @@ func pullDaemonlessImage(sc types.SystemContext, store storage.Store, imageName 
 		return fmt.Errorf("error parsing image name %s: %v", ref, err)
 	}
 	if ref.Registry == "" {
+		// Due to https://github.com/containers/image/issues/572, buildah will default
+		// the registry to docker.io. If docker.io is not in the registry search path
+		// (set in /etc/registries.conf), buildah may be able to pull the image,
+		// but will not be able to use it when it runs build commands.
 		glog.V(2).Infof("defaulting registry to docker.io for image %s", imageName)
-		ref.Registry = "docker.io"
 	}
 
 	if authConfig.Username != "" && authConfig.Password != "" {
-		glog.V(2).Infof("Setting authentication for registry %q for %q.", ref.Registry, imageName)
-		if err := config.SetAuthentication(&systemContext, ref.Registry, authConfig.Username, authConfig.Password); err != nil {
-			return err
+		if ref.Registry == "" {
+			glog.V(0).Infof("WARNING: authentication was provided, but no registry was specified for image %s.", imageName)
+		} else {
+			glog.V(2).Infof("Setting authentication for registry %q for %q.", ref.Registry, imageName)
+			if err := config.SetAuthentication(&systemContext, ref.Registry, authConfig.Username, authConfig.Password); err != nil {
+				return err
+			}
 		}
 	}
 


mtrmac:
I can’t see how this is even moving us closer to search working: `docker.AuthConfiguration` only contains an username+password, so it fundamentally can’t correctly support the search semantics, in which case multiple different authenticated registries may be present, and the code must not send credentials intended for one registry to other registries.

Ultimately, the data comes from https://github.com/kubernetes/kubernetes/blob/a780dd78d30616c5bbf254244013d784a73c5d39/pkg/credentialprovider/keyring.go#L275-L280 , which hard-codes the `docker.io[/library]` normalization as well.

Note that the same credential provider is used in the kubelet, so hostname-unqualied image names in pod specs have never really worked. Given the ambiguity about which image is run (which breaks signatures, and is generally risky anyway), and that the ultimate users of OpenShift need to qualify images in pod specs anyway, I would strongly recommend being consistent about it and not trying to support search at all.

If you wanted to completely reject unqualified names and to force users to always use a host name, that would be fine with me (though, maybe, impractical); otherwise, I think the `docker.io` default is preferable to search (which exposes the user to network attackers, who can make the first registry in the search unavailable to redirect the user to use an entirely different, quite possibly unexpected, image).

pkg/build/builder/daemonless.go
```diff
@@ -51,14 +51,21 @@ func pullDaemonlessImage(sc types.SystemContext, store storage.Store, imageName
 		return fmt.Errorf("error parsing image name %s: %v", ref, err)
 	}
 	if ref.Registry == "" {
+		// Due to https://github.com/containers/image/issues/572, buildah will default
+		// the registry to docker.io. If docker.io is not in the registry search path
+		// (set in /etc/registries.conf), buildah may be able to pull the image,
+		// but will not be able to use it when it runs build commands.
 		glog.V(2).Infof("defaulting registry to docker.io for image %s", imageName)
```
bparees: i don't think we should print this message.  at best you might want to log a "no registry specified" message, but we have no idea what the defaulting behavior will be.

pkg/build/builder/daemonless.go
```diff
@@ -51,14 +51,21 @@ func pullDaemonlessImage(sc types.SystemContext, store storage.Store, imageName
...
-		glog.V(2).Infof("Setting authentication for registry %q for %q.", ref.Registry, imageName)
-		if err := config.SetAuthentication(&systemContext, ref.Registry, authConfig.Username, authConfig.Password); err != nil {
-			return err
+		if ref.Registry == "" {
+			glog.V(0).Infof("WARNING: authentication was provided, but no registry was specified for image %s.", imageName)
+		} else {
+			glog.V(2).Infof("Setting authentication for registry %q for %q.", ref.Registry, imageName)
+			if err := config.SetAuthentication(&systemContext, ref.Registry, authConfig.Username, authConfig.Password); err != nil {
+				return err
+			}
```
bparees: Seems like we need to be able to pass the available auths to buildah and buildah should pick which auth (if any) to use as it iterates through the search list.

bparees:
@mtrmac we are trying to retain compatibility with 3.11 behavior in which admins could setup search lists and pass auth for one or more registries, and docker would (I assume) try each registry in order, using the auth for that registry if available.

It sounds like buildah does not support anything close to that model of behavior, in which case I would tend to agree that we could punt on supporting search at all because we can't do it properly.

We *must* default to docker.io at a minimum, however, because users absolutely have dockerfiles like"

```
FROM busybox
RUN foo
```

and we cannot break those users.

Let's talk tomorrow about precisely what buildah does/does not do 
1) when asked to pull an image
2) when pulling an image referenced in a dockerfile

/cc @derekwaynecarr since you were involved in the discussions about what to do for search.

pkg/build/builder/daemonless.go
adambkaplan: I kept this in the event https://github.com/openshift/origin/pull/21864 lands - tests there explicitly look for this message.

bparees: ok, we'll probably need a couple more turns of the crank (but let's let 21864 land as is) because ultimately i'd want to get rid of this.

adambkaplan:
> `docker.AuthConfiguration` only contains an username+password, so it fundamentally can’t correctly support the search semantics

@mtrmac the docker library we're using does include a `ServerAddress` field in `AuthConfiguration`, and as I understand it this is the registry host. This ultimately comes from upstream kube's `BasicDockerKeyring` - see https://github.com/openshift/builder/blob/master/pkg/build/builder/cmd/dockercfg/cfg.go#L63-L75.

So if we do have a populated `ServerAddress` field, will we be able to restore search semantics?

Another question related to this code path - does the docker config allow the server address key to be empty, and if so does this imply the credential is for `docker.io`?

adambkaplan:
@nalind how does auth work with buildah if we pass it a multistage dockerfile - particularly with multiple repos specified? Does it use docker semantics to locate the auth configs?

ex:
```
FROM quay.io/baseimage as build
...
FROM registry.redhat.io/runimage as run
...
```

pkg/build/builder/daemonless.go
adambkaplan: @nalind are there better ways of configuring auth for buildah? Can we point buildah to the right dockercfg JSON file(s) as we do in https://github.com/openshift/builder/blob/master/pkg/build/builder/cmd/dockercfg/cfg.go#L51-L61 ?

adambkaplan:
> Ultimately, the data comes from https://github.com/kubernetes/kubernetes/blob/a780dd78d30616c5bbf254244013d784a73c5d39/pkg/credentialprovider/keyring.go#L275-L280 , which hard-codes the `docker.io[/library]` normalization as well.

This IMO reinforces the need to keep `docker.io` as the only entry on the search list in the near term.

mtrmac:
> @mtrmac we are trying to retain compatibility with 3.11 behavior in which admins could setup search lists and pass auth for one or more registries, and docker would (I assume) try each registry in order, using the auth for that registry if available.

Looking at the 3.11 (and 3.10) code, it also always used only a single `AuthConfiguration` object, not a map, and typically did its own extraction from a docker.cfg via `…/credentialprovider` AFAICS (even in that one case of `docker.BuildImageOptions.AuthConfigs`, they end up only being a source for `BasicDockerKeyring.Lookup`).

In the Docker Engine API, the client is responsible for parsing the config file (and the upstream API does not allow passing multiple credentials at all — only the projectatomic/docker variant can do do that), either I have missed some code path (quite possible), or the only other way of passing credentials for multiple registries could possibly work is by executing an external `docker build` process.

Overall, it seems to me that _when passing credentials_ this has just never worked correctly. (It might have worked _incorrectly_, e.g. by parsing credentials for `docker.io` and handing them off the to a search code, with no control over which registries would see those credentials — which happened to be first/only `docker.io` in tests.)

> It sounds like buildah does not support anything close to that model of behavior, in which case I would tend to agree that we could punt on supporting search at all because we can't do it properly.

buildah the command does support search, at least on some code paths, and _if_ using such a path, ultimately passing a path to a complete `.docker/config.json` file via `c/image/types.SystemContext.AuthFilePath` should make it possible to pass multiple credentials.  (There isn’t currently a way to pass a `map[hostname]credentials`.) But it seems that the search code paths are not, or not always, use by builder.

> We _must_ default to docker.io at a minimum, however, because users absolutely have dockerfiles like:
> 
> ```
> FROM busybox
> RUN foo
> ```
> and we cannot break those users.

Sure.  It would be plausible, I think, to support search but to refuse to send _any_ credentials in that case, like this PR does (i.e. only to ever use search for anonymous access) — but this PR only does the credentials change, not the search.

---

Still, search is a very bad idea WRT security, and I’d much prefer just killing it if at all possible.

mtrmac:
> > `docker.AuthConfiguration` only contains an username+password, so it fundamentally can’t correctly support the search semantics
> 
> @mtrmac the docker library we're using does include a `ServerAddress` field in `AuthConfiguration`, and as I understand it this is the registry host.

Ah, I have missed that. Still, overall… see below.

> This ultimately comes from upstream kube's `BasicDockerKeyring` - see https://github.com/openshift/builder/blob/master/pkg/build/builder/cmd/dockercfg/cfg.go#L63-L75.

In that code path, it’s too late - `BasicDockerKeyring.Lookup` has already normalized the input to `docker.io` , and you can see that the code only returns a _single_ credential value.

Also note how `https://github.com/kubernetes/kubernetes/blob/a3ccea9d8743f2ff82e41b6c2af6dc2c41dc7b10/pkg/credentialprovider/keyring.go#L84` discards the `ServerAddress` value, if it ever existed, and how a typical `.docker/config.json` does not include the field at all, so it isn’t set by unmarshaling.

---

OTOH https://github.com/openshift/builder/blob/918f4590f10bc644bd0e2d8a481469410bb825d6/pkg/build/builder/daemonless.go#L102-L112 does seem to support credentials for multiple registries, correctly from a quick glance.

(It’s working too hard, FWIW: AFAICS both sources of the value, https://github.com/openshift/builder/blob/c6a4315f96e7af1777dee8178c5c372e08fd97c6/pkg/build/builder/dockerutil.go#L124 and https://github.com/openshift/builder/blob/8967e7a56f9df7643661ede68b52f2a4e6745090/pkg/build/builder/sti.go#L425 , start with a config file — just pointing `c/image/types.SystemContext.AuthFilePath` to it should be much simpler than parsing it and creating a new `/tmp/config.json`.)

---

> So if we do have a populated `ServerAddress` field, will we be able to restore search semantics?

Authentication is really the canary showing that this never worked, and I’d rather punt on this insecure design than make it work _now_. Of course my preferences don’t have to prevail :)

It’s not really the `ServerAddress` field at all — it’s the whole _concept_ of needing to pass a `map[hostname]credentials` to the builder, all the way from the unprivileged users’ build configurations through the builder images to the search code somewhere in buildah ultimately to c/image. If I understand the code (which I have never seen before) right, including things like https://github.com/openshift/origin/blob/9eb089df55d70e03ee888377eb22f9c8918adda9/pkg/build/controller/build/build_controller.go#L586 .

_Really, authentication + search could never have worked right._

As for _specifically_ the `ServerAddress` field, I can’t quite see the point of it - the `.docker/config.json` file is essentially a `map[host]docker/docker/api/types.AuthConfig`, and if `AuthConfig.ServerAddress` is ever used, it is set (e.g. in https://github.com/projectatomic/docker/blob/1185cfdf0469a0434e9227c973f8783412237b82/cliconfig/configfile/file.go#L70 ) from the key of that map when parsing the file.

---

> Another question related to this code path - does the docker config allow the server address key to be empty

In the config file, it’s the common case. In this particular code path, I don’t know…

> and if so does this imply the credential is for `docker.io`?

Typically if only one credential set is provided, it means “for the current image” and it’s up to the caller to match them.

pkg/build/builder/daemonless.go
mtrmac: This should be possible by setting `c/image/types.SystemContext.AuthFilePath`, and making sure `….SystemContext.DockerAuthConfig` is `nil`.

adambkaplan:
Thanks, @mtrmac for your great input. From the OpenShift perspective, I think we're in agreement this PR is not the way to go, and that we should remove search as a configurable option. I'll be opening separate PRs to get the ball rolling on that.

PR #43, opened 2019-02-13
Process $DROP_CAPS

nalind:
Process $DROP_CAPS (example value: "KILL,MKNOD,SETGID,SETUID,SYS_CHROOT") by splitting it at commas, upper-casing it, prefixing each value with "CAP_" if it doesn't already start with that prefix, and setting it in the `DropCapabilities` list in our `BuildOptions`.

### DIFF
--- a/pkg/build/builder/daemonless.go
+++ b/pkg/build/builder/daemonless.go
@@ -24,8 +24,26 @@ import (
 	"golang.org/x/sys/unix"
 
 	buildapiv1 "github.com/openshift/api/build/v1"
+	builderutil "github.com/openshift/builder/pkg/build/builder/util"
 	"github.com/openshift/library-go/pkg/image/reference"
 )
+
+// The build controller doesn't expect the CAP_ prefix to be used in the
+// entries in the list in the environment, but our runtime configuration
+// expects it to be provided, so massage the values into a suitabe list.
+func dropCapabilities() []string {
+	var dropCapabilities []string
+	if dropCaps, ok := os.LookupEnv(builderutil.DropCapabilities); ok && dropCaps != "" {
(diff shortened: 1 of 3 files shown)


bparees:
rather than wiring this though as env vars (given that the env vars are hardcoded in the origin build controller strategy impl), I think it makes more sense to just hardcode the caps to drop here in the builder image?  (and remove the logic in openshift/origin that sets the env vars, since I guess nothing is currently consuming them?)

/cc @adambkaplan 

In any case I guess none of this actually drops capabilities on the builder pod itself, which is what we really need to be doing if we want to tighten up the security of our privileged builder pod.

nalind:
Yes, this sets us up to drop capabilities when we're handling RUN instructions, but the builder process itself still has everything that we granted to the builder container.

rhatdan:
@adambkaplan PTAL

pkg/build/builder/daemonless.go
```diff
@@ -24,9 +24,27 @@ import (
...
+		for i := range dropCapabilities {
+			dropCapabilities[i] = strings.ToUpper(dropCapabilities[i])
+			if !strings.HasPrefix(dropCapabilities[i], "CAP_") {
+				dropCapabilities[i] = "CAP_" + dropCapabilities[i]
+			}
+		}
+	}
+	return dropCapabilities
+}
+
```
adambkaplan: @nalind unit tests for this?

adambkaplan:
hrm...also looks like upgrade tests aren't running on this repo. Need to check this is required.

nalind:
Added a unit test for the `dropCapabilities()` function, which parses the environment variable and returns a slice of capability names.

nalind:
Thanks, squashed the commits.

adambkaplan:
this will get stuck in tide until e2e-aws-builds is green again.

PR #46, opened 2019-03-01
S2IBuilder.pullImage: retry pulls

### LINKED ISSUE #45: retry image pulls
we should be retrying image pulls in the same way that we retry image pushes, so we do not fail builds on network flakes.

nalind:
Wrap S2IBuilder.pullImage's attempts to pull an image with retryImageAction(), to match its push logic and the pull and push logic that we use for builds that use the Docker strategy.

This should fix #45.

### DIFF
--- a/pkg/build/builder/sti.go
+++ b/pkg/build/builder/sti.go
@@ -451,7 +451,9 @@ func (s *S2IBuilder) pullImage(name string, authConfig dockerclient.AuthConfigur
 		options.Repository = name
 	}
 
-	return s.dockerClient.PullImage(options, authConfig)
+	return retryImageAction("Pull", func() (pullErr error) {
+		return s.dockerClient.PullImage(options, authConfig)
+	})
 }
 
 func (s *S2IBuilder) buildImage(optimization buildapiv1.ImageOptimizationPolicy, opts dockerclient.BuildImageOptions) error {


pkg/build/builder/sti.go
```diff
@@ -451,7 +451,9 @@ func (s *S2IBuilder) pullImage(name string, authConfig dockerclient.AuthConfigur
 		options.Repository = name
 	}
 
-	return s.dockerClient.PullImage(options, authConfig)
+	return retryImageAction("Pull", func() (pullErr error) {
+		return s.dockerClient.PullImage(options, authConfig)
+	})
```
bparees: don't we also do this in docker builder?

nalind: Yes, this is what DockerBuilder does.

bparees: also extractSourceFromImage in source.go pulls images indirectly (it calls Mount which i assume pulls the image) so we might need to retry the Mount if it fails due to a pull failure also.

bparees: oh I see, it was only missing in the s2i path.  got it.

Still wondering if we need to do something in the Mount flow when we're extracting content from other images?

bparees:
i'm gonna call this lgtm since it addresses the main issue, but if there's something that can/should be done for Mount, let's do a follow up.  thanks @nalind 

/lgtm

pkg/build/builder/sti.go
nalind: Oh, I see what you mean now.  `extractSourceFromImage()` depends on `buildah.NewBuilder()`  to implicitly pull the image that we want to extract content from, so it wasn't as obvious that we weren't retrying there.

smarterclayton:
https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_builder/46/pull-ci-openshift-builder-master-e2e-aws-builds/75/#openshift-tests-featurebuildsserialslow-alter-builds-via-cluster-configuration-registries-config-context-should-allow-registries-to-be-blacklisted-suiteopenshift hasn't passed in the last 5 runs on this - is that a known broken test (why isn't it fixed), a super flaky test (same), or something in this PR?

bparees:
@smarterclayton it's not this PR, it's also failing in this PR which just touches an owner file:
https://github.com/openshift/builder/pull/48

It's also not known to be flaky.  Looking at the failures, it looks like the failures are being caused by the openshift apiserver being unavailable.  We get either:

1) errors that the builds resource does not exist
or
2) errors that the api server is unavailable (presumably the issue I mentioned to you on thursday or friday about api servers not being removed from endpoints before they start rejecting requests).

Does the [SMOKE] suite run before all e2es?  Or is there something we need to do to ensure cluster stability before e2e-aws-builds gets kicked off?

bparees:
also any idea why /override is not doing what i would expect it to?

bparees:
(and none of this was known to be broken before friday which is why none of it is fixed).

bparees:
```
Mar  3 05:41:31.273: INFO: error listing builds: the server is currently unable to handle the request (get builds.build.openshift.io)
```

I don't know what makes this particular test more vulnerable to that issue (I don't believe anything it's doing should be triggering an api server restart, but it does touch the image cluster config object which the openshift apiserver watches, so perhaps there is a restart happening there that shouldn't be).

bparees:
we've also seen auth failures listing builds:

```
Mar  2 17:14:33.745: INFO: error listing builds: Unauthorized
Mar  2 17:14:33.745: INFO: WaitForABuild returning with error: Unauthorized
```

so basically all these things need to be opened as bugs against auth + master teams, but i was waiting until we can do a proper investigation on monday and chase people.

bparees:
last thought:  i'm inclined to suspect the change to the CVO to not wait for progressing=false if a version is published may be what introduced this instability (by exposing us to the issues w/ the apiserver returning 503s when quiescing).  So if you want a quick fix i'd like to see that reverted until the control plane handles quiescing in a way that doesn't require clients to tolerate/retry 503s (because adding that tolerance in all our tests is going to be a massive undertaking).

bparees:
more interesting data from this run:
https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_builder/46/pull-ci-openshift-builder-master-e2e-aws-builds/71/

```
           "lastTimestamp": "2019-03-02T17:14:48Z",
            "message": "Status for operator openshift-apiserver changed: Available changed from True to False (\"Available: v1.apps.openshift.io is not ready: 503\\nAvailable: v1.project.openshift.io is not ready: 503\")",
```

and

```
           "lastTimestamp": "2019-03-02T17:26:40Z",
            "message": "Status for operator openshift-apiserver changed: Available message changed from \"Available: v1.build.openshift.io is not ready: 503\\nAvailable: v1.oauth.openshift.io is not ready: 503\\nAvailable: v1.project.openshift.io is not ready: 503\\nAvailable: v1.route.openshift.io is not ready: 503\" to \"Available: v1.apps.openshift.io is not ready: 503\\nAvailable: v1.oauth.openshift.io is not ready: 503\\nAvailable: v1.route.openshift.io is not ready: 503\"",
```

PR #47, opened 2019-03-01
Bug 1682978: Report builder version

adambkaplan:
* Burn in version using git tags
* Log binary version if verbose logging is enabled

Fixes [Bug 1682978](https://bugzilla.redhat.com/show_bug.cgi?id=1682978)


bparees:
i've never fully understood all this scripting, but is there a reason we can't just do what @dmage did in the image registry operator?

https://github.com/openshift/cluster-image-registry-operator/blob/master/hack/build/build.sh#L17-L18

adambkaplan:
Existing code uses kube's `version.Info` object, hence the semver parsing. I started with the origin scripts and pared them down to only the things we care about. Origin's needs are obviously much more complex than ours.

pkg/build/builder/cmd/builder.go
```diff
@@ -7,29 +7,27 @@ import (
 	"os"
 	"path/filepath"
 
-	istorage "github.com/containers/image/storage"
-	"github.com/containers/image/types"
-	"github.com/containers/storage"
-	realglog "github.com/golang/glog"
```
bparees: do you understand the implications of this change?  There's a reason we were distinguishing between the glog package and the glog variable before...

adambkaplan: This was meant to clean up the imports.

pkg/build/builder/cmd/builder.go
```diff
@@ -318,7 +317,6 @@ func runBuild(out io.Writer, builder builder) error {
 
 // RunDockerBuild creates a docker builder and runs its build
 func RunDockerBuild(out io.Writer) error {
-	realglog.V(5)
```
adambkaplan: @bparees This got rid of the `realglog` import - did you intend to keep this?

pkg/build/builder/cmd/builder.go
```diff
@@ -372,7 +370,6 @@ func RunGitClone(out io.Writer) error {
 // and also adds some env and label values to the dockerfile based on
 // the build information.
 func RunManageDockerfile(out io.Writer) error {
-	serviceability.InitLogrus("DEBUG")
```
adambkaplan: @bparees ditto here - was this intended to be kept when you fixed the debug logging for buildah?

pkg/build/builder/cmd/builder.go
adambkaplan: My understanding is that this is related to how we handle buildah logs. The `glog` variable is a wrapper which lets us override the effective log level from buildah.

pkg/build/builder/cmd/builder.go
bparees: hm.  I don't think so, at least i can't think of why we'd want it like this.

pkg/build/builder/cmd/builder.go
bparees: hm.  i have no idea what that was doing, so i guess not.....

pkg/build/builder/cmd/builder.go
bparees: no, the loglevel in buildah is handled elsewhere.  the glog wrapper is, if i recall correctly, used to avoid printing timestamps unless debug level logging is enabled.

Anyway based on your comments it makes sense now that it's going away here.

adambkaplan:
/hold

aws builds red

adambkaplan:
/hold cancel

e2e-aws-builds should be green again

PR #50, opened 2019-03-08
Drop unused copyImageSource() function

nalind:
Remove copyImageSource(), and the DownloadFromContainer() interface member and its daemonless implementation, which we're not using.

### DIFF
--- a/pkg/build/builder/daemonless.go
+++ b/pkg/build/builder/daemonless.go
@@ -5,9 +5,7 @@ package builder
 import (
 	"context"
 	"fmt"
-	"io"
 	"os"
-	"path/filepath"
 	"strings"
 	"time"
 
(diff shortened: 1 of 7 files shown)


nalind:
Also removing `DockerBuilder.tar`, which we're not using.

adambkaplan:
Unrelated - ecosystem test failed because buildah took over 10 minutes to pull an image due to a pull timeout. @nalind  are you aware of any timeout parameters we may be able to tune.

bparees:
> Unrelated - ecosystem test failed because buildah took over 10 minutes to pull an image due to a pull timeout.

sometimes the registry we're pulling from is just slow/flaky, i wouldn't worry based on one failure..but it might be worth confirming that when the test passes, the it is not exceptionally slow (ie if we're consistently taking 9 minutes, we're at high risk of flaking.  if we're consistently at 2-3 minutes, it's ok assuming we consider 2-3 mins reasonable for pulling the image).

PR #51, opened 2019-03-11
Switch back to the openshift docker fork

nalind:
Switch from nalind's fork of github.com/moby/moby to openshift's.  Update github.com/containers/storage to v1.11.


glide.yaml
```diff
@@ -4,9 +4,8 @@ import:
   subpackages:
   - reference
 - package: github.com/docker/docker
-  repo:    https://github.com/nalind/docker.git
-  version: master
-
+  repo:    https://github.com/openshift/moby-moby
+  version: openshift-3.10-docker-b68221c
```
bparees: is there a reason not to move up to a newer commit?  looks like moby-moby has moved significantly beyond this level.

nalind: None specifically, but I thought this was the version we were using before we switched away from the openshift fork.

bparees: (it seems like we should create a 4.0 branch in moby-moby and add the carries to it)

bparees: we can merge this as a good first step but it probably makes sense to create a 4.0 branch at some point and move to that.

nalind: Sounds good to me.

PR #54, opened 2019-03-19
Ignore .vscode

adambkaplan:
/assign @gabemontero

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 openshift-builder
+.vscode/
 


coreydaley:
Maybe just a personal preference, but .vscode should be in your .gitignore_global (since it is an editor that you use) and is not really specific to either this project, or this type of project, such as ignoring a build directory or object files, etc.

gabemontero:
I think @coreydaley makes a good point

coreydaley:
@bparees showed me that there is precedent in other openshift repositories for this kind of behavior, and I will go along with whatever is decided, but I do oppose it.

bparees:
to be clear i wasn't illustrating the precedent to say "so we should do it here too.", i think @coreydaley makes a good point that there is a better solution here.  Was just observing that this PR is in good company :)

adambkaplan:
going to update my globals now...

PR #56, opened 2019-03-20
Fix empty name in status update

adambkaplan:
/assign @gabemontero 

fyi @smarterclayton

### DIFF
--- a/pkg/build/builder/common.go
+++ b/pkg/build/builder/common.go
@@ -252,6 +252,12 @@ func HandleBuildStatusUpdate(build *buildapiv1.Build, client buildclientv1.Build
 		if latestBuild == nil {
 			latestBuild, err = client.Get(build.Name, metav1.GetOptions{})
 			if err != nil {
+				latestBuild = nil
+				return false, nil
+			}
+			if latestBuild.Name == "" {
+				latestBuild = nil
+				err = fmt.Errorf("latest version of build %s is empty", build.Name)
 				return false, nil
 			}
 		}


pkg/build/builder/common.go
```diff
@@ -266,7 +266,7 @@ func HandleBuildStatusUpdate(build *buildapiv1.Build, client buildclientv1.Build
 		latestBuild.Status.Output.To = build.Status.Output.To
 		latestBuild.Status.Stages = timing.AppendStageAndStepInfo(latestBuild.Status.Stages, build.Status.Stages)
 
-		_, err = client.UpdateDetails(latestBuild.Name, latestBuild)
+		_, err = client.UpdateDetails(build.Name, latestBuild)
```
adambkaplan: Observed on CI:

```
error: Unable to update build status: resource name may not be empty
```

It's possible that the `latestBuild` object returned is an empty struct and no error is reported. This message is likely masking other errors that would have occurred (image pull timeouts were also observed in the same issue).

pkg/build/builder/common.go
```diff
@@ -252,6 +252,12 @@ func HandleBuildStatusUpdate(build *buildapiv1.Build, client buildclientv1.Build
 		if latestBuild == nil {
 			latestBuild, err = client.Get(build.Name, metav1.GetOptions{})
 			if err != nil {
+				latestBuild = nil
+				return false, nil
+			}
+			if latestBuild.Name == "" {
```
adambkaplan: If for whatever reason we don't get a Name, I think it's safe to assume the rest of the object is empty or invalid. In either case we should re-fetch the build object.

adambkaplan:
/retest

known flakes reported by build cop

adambkaplan:
/retest

ci flake

adambkaplan:
@gabemontero bump

gabemontero:
(though a unit test wouldn't hurt ... decided not to block on that)

PR #59, opened 2019-04-02
Bug 1698152: fix image source extraction with trailing "/."

adambkaplan:
* Fix copy behavior if the image source path ends with "/."
* Add unit tests for copying regular files and symlinks

/assign @coreydaley 

/cc @bparees 

fyi @openshift/sig-developer-experience

### DIFF
--- a/pkg/build/builder/source_test.go
+++ b/pkg/build/builder/source_test.go
@@ -304,3 +304,189 @@ func TestCloneFromBranch(t *testing.T) {
 		}
 	}
 }
+
+func TestCopyImageSourceFromFilesystem(t *testing.T) {
+
+	testCases := []struct {
+		name        string
+		testFiles   map[string]string
+		testLinks   map[string]string
+		copyPath    string
+		destination string
+		verifyFiles map[string]string
+		verifyLinks map[string]string
+	}{
+		{
+			name: "single file",
+			testFiles: map[string]string{
+				"hello.txt": "Hello world!",
+			},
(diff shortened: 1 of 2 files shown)


bparees:
@adambkaplan whew, i was worried this might be deeper in buildah's image mounting/copying logic.

bparees:
(also i can't wait for someone to complain that we are not preserving their symlinks...  i'm pretty sure we've gone back and forth on this behavior in a variety of our file copying/moving codepaths, multiple times)

adambkaplan:
/hold

Trying to make sure this actually works via e2e tests

adambkaplan:
@bparees your fears may have been justified. See https://github.com/openshift/origin/pull/22456#issuecomment-479285125

pkg/build/builder/source_test.go
```diff
@@ -304,3 +304,72 @@ func TestCloneFromBranch(t *testing.T) {
 		}
 	}
 }
+
+func TestCopyImageSourceFromFilesystem(t *testing.T) {
```
coreydaley: This test should create the symlink in the src directory to a file in the link directory, then copy the src directory to the destination directory.  That would seem to follow more along the lines of what we would actually do?  Might just be me, but it seemed confusing the way it is done now, trying to figure out why things were being created/symlinked to other places that didn't make sense.

adambkaplan: @coreydaley test has been updated to be table-driven. Hopefully it's clearer.

PR #60, opened 2019-04-04
Do not print credentials into build log

legionus:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1694861
Tests will be in a follow up PR for the origin repo,

/assign @adambkaplan

### DIFF
--- a/pkg/build/builder/source.go
+++ b/pkg/build/builder/source.go
@@ -387,7 +387,7 @@ func extractSourceFromImage(ctx context.Context, dockerClient DockerClient, stor
 
 	if auths != nil {
 		for registry, ac := range auths.Configs {
-			glog.Infof("Setting authentication for registry %q using %#v", registry, ac)
+			glog.V(5).Infof("Setting authentication for registry %q using %q.", registry, ac.ServerAddress)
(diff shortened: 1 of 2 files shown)


pkg/build/builder/source.go
```diff
@@ -387,7 +387,7 @@ func extractSourceFromImage(ctx context.Context, dockerClient DockerClient, stor
 
 	if auths != nil {
 		for registry, ac := range auths.Configs {
-			glog.Infof("Setting authentication for registry %q using %#v", registry, ac)
+			glog.Infof("Setting authentication for registry %q using %q.", registry, ac.ServerAddress)
```
adambkaplan: Log level should be V(5) for this (debug). - `glog.V(5).Infof(...)`
There are similar log statements to this which should also be V(5):

https://github.com/openshift/builder/blob/master/pkg/build/builder/daemonless.go#L55-L60
https://github.com/openshift/builder/blob/master/pkg/build/builder/daemonless.go#L102-L109

legionus: @adambkaplan fixed

adambkaplan:
/retest

Hitting the Unauthorized bug

PR #61, opened 2019-04-05
Bug 1696756: Fix Assemble User in S2I builds

adambkaplan:
S2I needs to know the AssembleUser before it generates
the Dockerfile, otherwise a hard-coded default is used.

Replicate the logic in source-to-image:
1. Default to the most recent USER in the image
2. Override if the io.openshift.s2i.assemble-user label exists

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1696756

### DIFF
--- a/pkg/build/builder/sti.go
+++ b/pkg/build/builder/sti.go
@@ -254,15 +254,18 @@ func (s *S2IBuilder) Build() error {
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


adambkaplan:
/assign @gabemontero 

Extended test to check this is forthcoming in origin

adambkaplan:
/retest

unauthorized flake

PR #66, opened 2019-04-29
retryImageAction: strip error wrappings for ErrUnauthorizedForCredentials

nalind:
If the error we're reporting for a push or pull is an `ErrUnauthorizedForCredentials` error, strip out the context that other API layers have added, which tends to be a few copies of the image name, which we've previously logged, so that our error message looks like `After retrying %d times, pulling/pushing image still failed due to error: unable to retrieve auth token: invalid username/password`, per https://bugzilla.redhat.com/show_bug.cgi?id=1703262

### DIFF
--- a/pkg/build/builder/dockerutil.go
+++ b/pkg/build/builder/dockerutil.go
@@ -9,7 +9,9 @@ import (
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


adambkaplan:
note @nalind updated the title so we get the bug summary in our changelog

PR #67, opened 2019-04-29
Bugs 1695622,1699129: bump buildah to v1.8.2, pin most dependencies

nalind:
Update the version of buildah that we use to version 1.8.0, but downgrade the version of storage that we use to an older version.


nalind:
It looks like we broke something for builds that use post-commit hooks.

nalind:
Updated to be a bit newer than 1.8.0, closer to what I expect 1.8.1 to be, and pinning most other dependencies.

nalind:
Bumped to the commit ID that will be 1.8.1.

nalind:
Moved our buildah reference in glide.yaml to the v1.8.1 tag instead of the commit ID that it points to.  Pulled in some changes to the openshift branch of kubernetes.

nalind:
Bumped to 1.8.2.

pkg/build/builder/common.go
```diff
@@ -14,8 +14,8 @@ import (
 	"time"
 
 	"github.com/docker/distribution/reference"
-	dockercmd "github.com/docker/docker/builder/dockerfile/command"
-	"github.com/docker/docker/builder/dockerfile/parser"
+	dockercmd "github.com/openshift/imagebuilder/dockerfile/command"
+	"github.com/openshift/imagebuilder/dockerfile/parser"
```
adambkaplan: Why the change from docker/docker/builder to imagebuilder?

nalind: imagebuilder incorporated it to avoid versioning conflicts, so the type we get from `imagebuilder.ParseDockerfile()` no longer matched the argument type for local functions like `nodeHasFromRef()`.

adambkaplan:
/lgtm

Unclear why image-ecosystem tests flaked twice on the perl/dancer image.

PR #68, opened 2019-04-30
retryImageAction: strip error wrappings for MultiError members

nalind:
If the error we get back from attempting to push or pull an image is a multi-error, extend our change from #66 to process each of its component errors as we would have a single error.

### DIFF
--- a/pkg/build/builder/dockerutil.go
+++ b/pkg/build/builder/dockerutil.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	idocker "github.com/containers/image/docker"
+	"github.com/docker/distribution/registry/api/errcode"
 	docker "github.com/fsouza/go-dockerclient"
 	"github.com/pkg/errors"
 
@@ -37,6 +38,16 @@ type DockerClient interface {
 	TagImage(name string, opts docker.TagImageOptions) error
 }
 
+func unwrapUnauthorizedError(err error) error {
+	switch cause := errors.Cause(err); cause {
+	case idocker.ErrUnauthorizedForCredentials:
+		// strip off wrappers that mainly add the image name as their added context,
+		// which just duplicates information that we're already logging
+		return cause
+	}
+	return err
+}
+
 func retryImageAction(actionName string, action func() error) error {
 	var err error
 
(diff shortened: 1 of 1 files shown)


nalind:
Doh.  Forgot to mention the bug ID in the commit log message.

pkg/build/builder/dockerutil.go
```diff
@@ -49,11 +50,20 @@ func retryImageAction(actionName string, action func() error) error {
...
-		// strip off wrappers that mainly add the image name as their added context
-		err = errors.Cause(err)
+	processError := func(err error) error {
+		switch errors.Cause(err) {
+		case idocker.ErrUnauthorizedForCredentials:
+			// strip off wrappers that mainly add the image name as their added context
+			return errors.Cause(err)
+		}
+		return err
+	}
```
adambkaplan: Please make this a standalone function - could be useful in other contexts

nalind: Okay, changing both the check for the specific error and the iterating through MultiErrors.  If you only want the unwrap broken out, I'm happy to save that for `retryImageAction()`.

adambkaplan: I was thinking of just breaking out the unwrap. The multiError seems specific to the docker client.

nalind: Okay, I think that's got it.  The multiple-errors case turned out to be a different type than I'd thought.

rhatdan:
@adambkaplan Can we get this merged, seems to have gotten lost in the weeds.

rhatdan:
@gabemontero @bparees @adambkaplan PTAL

pkg/build/builder/dockerutil.go
```diff
@@ -49,12 +60,23 @@ func retryImageAction(actionName string, action func() error) error {
...
+		var unwrap bool
+		for i := range errs {
+			if err, ok := errs[i].(errcode.Error); ok {
+				if err.Code == errcode.ErrorCodeUnauthorized {
+					unwrap = true
+				}
+			}
+		}
+		if unwrap {
+			err = errors.Cause(err)
```
bparees: isn't this stomping the original group of errors?  so now you'll only log a single (unwrapped)error from the original group?

nalind: The scoping here is probably confusing - the `errors.Cause()` call on line 74 is removing additional context (that `errors.Wrapf()` would have added) from the `err` defined on line 52, when we detect that there's an authorization error in there somewhere.  If the result that first went into `errors.Wrapf()` is ultimately an `errcode.Errors`, that list is what `err` will be set to on line 74.  I'll change the name of the one we create on line 67.

bparees: sorry i'm still a little confused:
`if errs, ok := errors.Cause(err).(errcode.Errors); ok`

if that's true, then `err = errors.Cause(err)` is going to set err== something that is an array of errors (errcode.Errors)

then you're going to pass array of errors (interface errcode.Errors) down to unwrapUnauthorizedError

which then calls `cause := errors.Cause(err)`.  What does errors.Cause(err) return when err is an errcode.Errors?

I would have expected logic that iterates the array and unwraps each err within the array that needs unwrapping.

nalind: `errors.Cause()` looks at the error that `err` wraps, repeating until it reaches an error that isn't a wrapper (i.e., an object that doesn't provide a `Cause()` method), and `errcode.Errors` does not provide such a method, so if it's an `errcode.Errors`, and one of the errors in its list is an "unauthorized" error that we recognize, we're removing the layers of wrapping around the errors list.

bparees: ok.....so if action() returned an errors list or a wrapped errors list, you will scan the errors list, see if it contains an "unauthorized error", and then attempt to unwrap the list (leaving the list itself).

if that's the intent (as compared w/ unwrapping the unauthorized error specifically), then this lgtm

PR #69, opened 2019-05-02
Raise file and process limits before building an image

nalind:
Attempt to set the limit on open files and processes while we're building an image to 0x100000 (1048576), overriding the default open files limit that we get when generating a runtime configuration.


mrunalp:
@bparees @adambkaplan ptal

bparees:
@gabemontero seeing the same jenkins test failure here: https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_builder/69/pull-ci-openshift-builder-master-e2e-aws-builds/203

PR #70, opened 2019-05-07
Bug 1704767: S2I incremental build not working

coreydaley:
S2I incremental builds were failing due to not having the correct authorization
to pull the previously built image.

### DIFF
--- a/pkg/build/builder/sti.go
+++ b/pkg/build/builder/sti.go
@@ -241,8 +241,11 @@ func (s *S2IBuilder) Build() error {
 
 	if config.Incremental {
 		if s.build.Spec.Strategy.SourceStrategy.ForcePull || !isImagePresent(s.dockerClient, config.IncrementalFromTag) {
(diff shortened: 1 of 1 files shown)


pkg/build/builder/sti.go
```diff
@@ -241,8 +241,9 @@ func (s *S2IBuilder) Build() error {
 
 	if config.Incremental {
 		if s.build.Spec.Strategy.SourceStrategy.ForcePull || !isImagePresent(s.dockerClient, config.IncrementalFromTag) {
+			it, _ := dockercfg.NewHelper().GetDockerAuth(config.IncrementalFromTag, dockercfg.PullAuthType)
```
bparees: it's counterintuitive but this should probably be dockercfg.PushAuthType

because you are pulling from the push destination, which means you would be using the same creds you are using to push to that destination.

bparees: (it is probably working for you because the push + pull creds are the same, by default:  the SA's creds that can access the internal registry)

PR #74, opened 2019-05-16
Parse $BUILD_STORAGE_OPTIONS to set storage options

nalind:
Parse $BUILD_STORAGE_OPTIONS, if set, as a JSON string array, so that we can set storage options passed to us by the build controller.

### DIFF
--- a/pkg/build/builder/cmd/builder.go
+++ b/pkg/build/builder/cmd/builder.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"io"
 	"os"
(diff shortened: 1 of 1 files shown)


nalind:
This'll be useful for letting the build controller tell us about read-only image stores that it's providing in volumes, and whether or not to use an external binary like fuse-overlayfs to handle mounting overlay filesystems.  JSON because I didn't want to invent my own quoting rules, though if there's a different precedent for how to pass lists of strings from the build controller to the builder that I didn't spot, I'd lean toward changing it to that.

rhatdan:
Ok my latest patch works similar to this.  Lets merge yours and then I can rebase mine.

LGTM

PR #75, opened 2019-05-16
push: log the image's canonical reference

nalind:
Log a canonical reference to the image that we push.  The buildah library will stop writing this message directly to stdout on its own starting with https://github.com/containers/buildah/pull/1596.

### DIFF
--- a/pkg/build/builder/daemonless.go
+++ b/pkg/build/builder/daemonless.go
@@ -12,6 +12,7 @@ import (
 	"github.com/containers/buildah"
 	"github.com/containers/buildah/imagebuildah"
 	"github.com/containers/buildah/util"
+	ireference "github.com/containers/image/docker/reference"
 	"github.com/containers/image/pkg/docker/config"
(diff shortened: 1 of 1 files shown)


nalind:
The messages before and after:
```
Successfully pushed //image-registry.openshift-image-registry.svc:5000/default/mydockertest:latest@sha256:48de4e39a144c768a7b60212fcb82a117c41f49b26f8fbe2a2d6cbd5c9f81040
```
```
Successfully pushed image-registry.openshift-image-registry.svc:5000/default/mydockertest@sha256:48de4e39a144c768a7b60212fcb82a117c41f49b26f8fbe2a2d6cbd5c9f81040
```

The patch currently drops the tag portion of the name that we're pushing, but we can add it back if it's useful to continue having it there.

adambkaplan:
@nalind I'd prefer to keep the tag name in. That lets me as a user know that the image was pushed to the right repo with the correct tag.

nalind:
Stopped stripping the tag from the pushed reference.

PR #76, opened 2019-05-23
Bug 1712245: bump github.com/containers/buildah and github.com/openshift/source-to-image

nalind:
Bump buildah to get a fix for how it recognizes which cached image to use, which it hadn't always been getting right since it stopped creating layers for every instruction, and to fix part of https://bugzilla.redhat.com/show_bug.cgi?id=1712245 by pulling in https://github.com/containers/buildah/pull/1607.  This pulls in `github.com/cyphar/filepath-securejoin` as a new dependency.


nalind:
Added a bump of source-to-image to pull in https://github.com/openshift/source-to-image/pull/969, which along with the buildah update, should fix https://bugzilla.redhat.com/show_bug.cgi?id=1712245.

nalind:
Bumped buildah to pick up https://github.com/containers/buildah/pull/1620.

adambkaplan:
/cherrypick release-4.1

Giving it a shot.

PR #79, opened 2019-06-24
Change containers to allow building without privilege.

rhatdan:
By turning on fuse-overlayfs, the container can run without the
--privileged flag.

Also setup shared directory to allow images to be shared into
the containers via the volume mounts from the host.

Signed-off-by: Daniel J Walsh <dwalsh@redhat.com>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,9 +5,8 @@ RUN hack/build.sh
 
 FROM registry.svc.ci.openshift.org/openshift/origin-v4.0:base
 RUN INSTALL_PKGS=" \
-      which tar wget hostname sysvinit-tools util-linux \
-      socat tree findutils lsof bind-utils \
-      git tar bsdtar \
+      bind-utils bsdtar findutils fuse-overlayfs git hostname lsof socat \
+      sysvinit-tools tar tree util-linux wget which \
       " && \
     yum install -y $INSTALL_PKGS && \
     yum clean all
@@ -15,7 +14,11 @@ COPY --from=builder /go/src/github.com/openshift/builder/openshift-builder /usr/
 COPY imagecontent/policy.json /etc/containers/
 COPY imagecontent/registries.conf /etc/containers/
 COPY imagecontent/storage.conf /etc/containers/
-RUN mkdir /var/cache/blobs
+RUN mkdir -p /var/cache/blobs \
+    /var/lib/shared/overlay-images \
+    /var/lib/shared/overlay-layers && \
+    touch /var/lib/shared/overlay-images/images.lock \
+    /var/lib/shared/overlay-layers/layers.lock
 
 RUN ln -s /usr/bin/openshift-builder /usr/bin/openshift-sti-build && \
     ln -s /usr/bin/openshift-builder /usr/bin/openshift-docker-build && \
--- a/Dockerfile-dev
+++ b/Dockerfile-dev
@@ -4,9 +4,8 @@ LABEL io.k8s.display-name="OpenShift Origin Builder" \
       io.openshift.tags="openshift,builder"
 
 RUN INSTALL_PKGS=" \
-      which tar wget hostname sysvinit-tools util-linux \
-      socat tree findutils lsof bind-utils \
-      git tar bsdtar \
+      bind-utils bsdtar findutils fuse-overlayfs git hostname lsof socat \
+      sysvinit-tools tar tree util-linux wget which \
       " && \
     yum install -y ${INSTALL_PKGS} && \
     yum clean all
@@ -14,7 +13,11 @@ RUN INSTALL_PKGS=" \
 COPY imagecontent/policy.json /etc/containers/
 COPY imagecontent/registries.conf /etc/containers/
 COPY imagecontent/storage.conf /etc/containers/
-RUN mkdir /var/cache/blobs
+RUN mkdir -p /var/cache/blobs \
+    /var/lib/shared/overlay-images \
+    /var/lib/shared/overlay-layers && \
+    touch /var/lib/shared/overlay-images/images.lock \
+    /var/lib/shared/overlay-layers/layers.lock
 
 COPY openshift-builder /usr/bin
 RUN ln -s /usr/bin/openshift-builder /usr/bin/openshift-sti-build && \
--- a/Dockerfile.rhel7
+++ b/Dockerfile.rhel7
@@ -5,9 +5,8 @@ RUN hack/build.sh
 
 FROM registry.svc.ci.openshift.org/ocp/4.0:base
 RUN INSTALL_PKGS=" \
-      which tar wget hostname sysvinit-tools util-linux \
-      socat tree findutils lsof bind-utils \
-      git tar bsdtar \
+      bind-utils bsdtar findutils git hostname lsof socat \
+      sysvinit-tools tar tree util-linux wget which \
       " && \
     yum install -y $INSTALL_PKGS && \
     yum clean all
@@ -15,7 +14,11 @@ COPY --from=builder /go/src/github.com/openshift/builder/openshift-builder /usr/
 COPY imagecontent/policy.json /etc/containers/
 COPY imagecontent/registries.conf /etc/containers/
 COPY imagecontent/storage.conf /etc/containers/
-RUN mkdir /var/cache/blobs
+RUN mkdir -p /var/cache/blobs \
+    /var/lib/shared/overlay-images \
+    /var/lib/shared/overlay-layers && \
+    touch /var/lib/shared/overlay-images/images.lock \
+    /var/lib/shared/overlay-layers/layers.lock
 
 RUN ln -s /usr/bin/openshift-builder /usr/bin/openshift-sti-build && \
     ln -s /usr/bin/openshift-builder /usr/bin/openshift-docker-build && \
--- a/imagecontent/storage.conf
+++ b/imagecontent/storage.conf
@@ -19,6 +19,7 @@ graphroot = "/var/lib/containers/storage"
 # AdditionalImageStores is used to pass paths to additional Read/Only image stores
 # Must be comma separated list.
 additionalimagestores = [
+"/var/lib/shared",
 ]
 
 # Size is used to set a maximum size of the container image.  Only supported by
@@ -28,9 +29,6 @@ size = ""
 # Path to an helper program to use for mounting the file system instead of mounting it
 # directly.
 #mount_program = "/usr/bin/fuse-overlayfs"
-
-# OverrideKernelCheck tells the driver to ignore kernel checks based on kernel version
-override_kernel_check = "false"
 
 # mountopt specifies comma separated list of extra mount options
 mountopt = ""
@@ -55,13 +53,6 @@ mountopt = ""
 #
 # remap-user = "storage"
 # remap-group = "storage"
-
-# If specified, use OSTree to deduplicate files with the overlay backend
-ostree_repo = ""
-
-# Set to skip a PRIVATE bind mount on the storage home directory.  Only supported by
-# certain container storage drivers
-skip_mount_home = "false"
 
 [storage.options.thinpool]
 # Storage Options for thinpool
@@ -114,7 +105,7 @@ skip_mount_home = "false"
 # mkfsarg = ""
 
 # use_deferred_removal marks devicemapper block device for deferred removal.
-# If the thinpool is in use when the driver attempts to remove it, the driver 
+# If the thinpool is in use when the driver attempts to remove it, the driver
 # tells the kernel to remove it as soon as possible. Note this does not free
 # up the disk space, use deferred deletion to fully remove the thinpool.
 # use_deferred_removal = "True"
--- a/pkg/build/builder/util/consts.go
+++ b/pkg/build/builder/util/consts.go
@@ -26,7 +26,7 @@ const (
 	StatusMessageFetchSourceFailed               = "Failed to fetch the input source."
 	StatusMessageInvalidContextDirectory         = "The supplied context directory does not exist."
 	StatusMessageCancelledBuild                  = "The build was cancelled by the user."
-	StatusMessageDockerBuildFailed               = "Docker build strategy has failed."
+	StatusMessageDockerBuildFailed               = "Dockerfile build strategy has failed."
 	StatusMessageBuildPodExists                  = "The pod for this build already exists and is older than the build."
 	StatusMessageNoBuildContainerStatus          = "The pod for this build has no container statuses indicating success or failure."
 	StatusMessageFailedContainer                 = "The pod for this build has at least one container with a non-zero exit status."


rhatdan:
@nalind @TomSweeneyRedHat PTAL

Dockerfile
```diff
@@ -6,7 +6,7 @@ RUN hack/build.sh
 FROM registry.svc.ci.openshift.org/openshift/origin-v4.0:base
 RUN INSTALL_PKGS=" \
       which tar wget hostname sysvinit-tools util-linux \
-      socat tree findutils lsof bind-utils \
+      socat tree findutils lsof bind-utils fuse-overlayfs\
```
nalind: The base image is derived from RHEL 7.6 at the moment.  Is fuse-overlayfs available there?

imagecontent/storage.conf
```diff
@@ -27,10 +28,7 @@ size = ""
 
 # Path to an helper program to use for mounting the file system instead of mounting it
 # directly.
-#mount_program = "/usr/bin/fuse-overlayfs"
-
-# OverrideKernelCheck tells the driver to ignore kernel checks based on kernel version
-override_kernel_check = "false"
+mount_program = "/usr/bin/fuse-overlayfs"
```
nalind: This forces us to use fuse-overlayfs even when we're running with privileges.

nalind: I'd prefer that we wire in a variable somewhere around [here](https://github.com/openshift/builder/blob/9277419505b020ca50c35ce1c045526413249c93/pkg/build/builder/cmd/builder.go#L118), that the build controller can set, that would let it tell us to use this when it's running us without privileges.

Dockerfile
rhatdan: Not sure, but I believe so.
@giuseppe Do you know?

imagecontent/storage.conf
rhatdan: We could do this, but fuse-overlay will work with --privilege as well.

rhatdan:
@nalind Is there a way to specify that we want --device /dev/fuse inside of containers?

nalind:
@rhatdan If the device is on the host, I think we can add a `HostPath` volume in the pod spec of type `HostPathCharDev` ("CharDevice") and mount it from there, somewhere near [the logic that sets the storage driver](https://github.com/openshift/openshift-controller-manager/blob/2e855b437b1d4fcbbd5fee46a64fe31d0f9128af/pkg/build/controller/strategy/util.go#L412).

If it's not on the node, I'm not sure if there's a place for it in the pod spec that we use to run the build, so we may have to try to include the device node in the image.

imagecontent/storage.conf
rhatdan: Added a check for BUILD_PRIVILEGED, and drop any GraphDriverOptions if set.

Dockerfile
giuseppe: AFAIK, fuse-overlayfs is not available on RHEL 7.6, since it is not usable there (at least not from a user namespace)

giuseppe: @lsm5 is fuse-overlayfs available for RHEL-7.6?  Do we plan to do it?

imagecontent/storage.conf
nalind: That would require the build controller to start setting the variable to continue to get the current behavior, which I'm not really in favor of, since updating the builder image before updating the controller would break builds.  A method for passing in driver options that the controller can use to set a `mount_program` would work, without imposing that ordering/upgrade requirement.

rhatdan: @nalind Now I pass in a "," list of options in the BUILD_STORAGE_OPTIONS environment variable?  Is this what you want?

Dockerfile
adambkaplan: @nalind I believe we have a tech debt card to base this on UBI, as we do for other OpenShift images

Dockerfile
```diff
@@ -15,7 +15,7 @@ COPY --from=builder /go/src/github.com/openshift/builder/openshift-builder /usr/
 COPY imagecontent/policy.json /etc/containers/
 COPY imagecontent/registries.conf /etc/containers/
 COPY imagecontent/storage.conf /etc/containers/
-RUN mkdir /var/cache/blobs
+RUN mkdir -p /var/cache/blobs /var/lib/shared/overlay-images /var/lib/shared/overlay-layers; touch /var/lib/shared/overlay-images/images.lock; touch /var/lib/shared/overlay-layers/layers.lock
```
adambkaplan: nit:
```suggestion
RUN mkdir -p /var/cache/blobs /var/lib/shared/overlay-images /var/lib/shared/overlay-layers && \
    touch /var/lib/shared/overlay-images/images.lock && \
    touch /var/lib/shared/overlay-layers/layers.lock
```

Dockerfile-dev
```diff
@@ -5,7 +5,7 @@ LABEL io.k8s.display-name="OpenShift Origin Builder" \
 
 RUN INSTALL_PKGS=" \
       which tar wget hostname sysvinit-tools util-linux \
-      socat tree findutils lsof bind-utils \
+      socat tree findutils lsof bind-utils fuse-overlayfs\
```
adambkaplan: nit - can we reorganize these to be in alpha order (looks like there are dupes)

Dockerfile-dev
```diff
@@ -14,7 +14,7 @@ RUN INSTALL_PKGS=" \
 COPY imagecontent/policy.json /etc/containers/
 COPY imagecontent/registries.conf /etc/containers/
 COPY imagecontent/storage.conf /etc/containers/
-RUN mkdir /var/cache/blobs
+RUN mkdir -p /var/cache/blobs /var/lib/shared/overlay-images /var/lib/shared/overlay-layers; touch /var/lib/shared/overlay-images/images.lock; touch /var/lib/shared/overlay-layers/layers.lock
```
adambkaplan: nit:
```suggestion
RUN mkdir -p /var/cache/blobs /var/lib/shared/overlay-images /var/lib/shared/overlay-layers && \
    touch /var/lib/shared/overlay-images/images.lock && \
    touch /var/lib/shared/overlay-layers/layers.lock
```

Dockerfile.rhel7
```diff
@@ -6,7 +6,7 @@ RUN hack/build.sh
 FROM registry.svc.ci.openshift.org/ocp/4.0:base
 RUN INSTALL_PKGS=" \
       which tar wget hostname sysvinit-tools util-linux \
-      socat tree findutils lsof bind-utils \
+      socat tree findutils lsof bind-utils fuse-overlayfs\
```
adambkaplan: ditto - reorganize in alpha order

Dockerfile.rhel7
```diff
@@ -15,7 +15,7 @@ COPY --from=builder /go/src/github.com/openshift/builder/openshift-builder /usr/
 COPY imagecontent/policy.json /etc/containers/
 COPY imagecontent/registries.conf /etc/containers/
 COPY imagecontent/storage.conf /etc/containers/
-RUN mkdir /var/cache/blobs
+RUN mkdir -p /var/cache/blobs /var/lib/shared/overlay-images /var/lib/shared/overlay-layers; touch /var/lib/shared/overlay-images/images.lock; touch /var/lib/shared/overlay-layers/layers.lock
```
adambkaplan: nit:
```suggestion
RUN mkdir -p /var/cache/blobs /var/lib/shared/overlay-images /var/lib/shared/overlay-layers && \
    touch /var/lib/shared/overlay-images/images.lock && \
    touch /var/lib/shared/overlay-layers/layers.lock
```

imagecontent/storage.conf
nalind: Options like "mountopt" include commas, so in the interest of not having to come up with an escaping scheme or having to change this later, I'd suggest expecting it to contain a JSON-encoded slice of strings that we unmarshal and use directly.  The value is set by the controller logic when it creates the pod spec that ultimately controls how a container is started using the image, and the controller already sets JSON-encoded values in environment variables for more complicated structures.

Dockerfile
rhatdan: Where is registry.svc.ci.openshift.org/openshift/origin-v4.0:base defined?  Can we update this to RHEL8/UBI8?

adambkaplan: Actually - this base image is replaced by CI, see https://github.com/openshift/release/blob/master/ci-operator/config/openshift/builder/openshift-builder-master.yaml#L2-L5

I believe this is in fact UBI based already (not sure if it's 7 or 8 - @bparees @stevekuznetsov can we verify this?)

adambkaplan: `Dockerfile-dev` is intended for contributor use - currently it's our public CentOS 7 image.

If "downstream" is building on top of UBI8, then I don't see an issue updating this to UBI8 as well.

bparees: it's ubi7 currently, as @adambkaplan notes we're based on a common openshift base so when it moves, we'll move automatically.

```
$ oc adm release info quay.io/openshift-release-dev/ocp-release:4.1.3 --pullspecs | grep builder
  docker-builder                                quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:883dbe5905ae08f05598fc82cd5e470f1f2c2aaac86e9976351fda3cdb924a51

$ docker run -it --entrypoint=/bin/sh quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:883dbe5905ae08f05598fc82cd5e470f1f2c2aaac86e9976351fda3cdb924a51 
Unable to find image 'quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:883dbe5905ae08f05598fc82cd5e470f1f2c2aaac86e9976351fda3cdb924a51' locally
sha256:883dbe5905ae08f05598fc82cd5e470f1f2c2aaac86e9976351fda3cdb924a51: Pulling from openshift-release-dev/ocp-v4.0-art-dev
18f0f6de4601: Already exists 
db1d55616933: Already exists 
25d3230a0c3e: Already exists 
f7a985d90b92: Pull complete 
Digest: sha256:883dbe5905ae08f05598fc82cd5e470f1f2c2aaac86e9976351fda3cdb924a51
Status: Downloaded newer image for quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:883dbe5905ae08f05598fc82cd5e470f1f2c2aaac86e9976351fda3cdb924a51

sh-4.2# cat /etc/redhat-release 
Red Hat Enterprise Linux Server release 7.6 (Maipo)

sh-4.2# uname -a
Linux ea33ab12ba9d 5.1.6-300.fc30.x86_64 #1 SMP Fri May 31 17:43:23 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux

sh-4.2# cd /etc/yum.repos.d/
sh-4.2# ls
redhat.repo  ubi.repo
```

rhatdan: Please use Podman for your examples :^)

rhatdan: Also could you see if you can install fuse-overlayfs, I don't think you can on RHEL7 based images.

bparees: I don't suppose you'd believe me if i told you i had it aliased.....

```
# yum install fuse-overlayfs
Loaded plugins: ovl, product-id, search-disabled-repos, subscription-manager
This system is not registered with an entitlement server. You can use subscription-manager to register.
ubi-7                                                              | 3.8 kB  00:00:00     
ubi-7-optional                                                     | 3.7 kB  00:00:00     
ubi-7-rhah                                                         | 3.7 kB  00:00:00     
ubi-7-rhscl                                                        | 3.8 kB  00:00:00     
(1/12): ubi-7/x86_64/updateinfo                                    |   92 B  00:00:00     
(2/12): ubi-7/x86_64/group                                         |  124 B  00:00:00     
(3/12): ubi-7/x86_64/primary_db                                    | 675 kB  00:00:00     
(4/12): ubi-7-optional/x86_64/group                                |  124 B  00:00:00     
(5/12): ubi-7-optional/x86_64/updateinfo                           |   92 B  00:00:00     
(6/12): ubi-7-optional/x86_64/primary_db                           |  13 kB  00:00:00     
(7/12): ubi-7-rhah/x86_64/group                                    |  124 B  00:00:00     
(8/12): ubi-7-rhah/x86_64/updateinfo                               |   92 B  00:00:00     
(9/12): ubi-7-rhah/x86_64/primary_db                               | 4.0 kB  00:00:00     
(10/12): ubi-7-rhscl/x86_64/group                                  |  124 B  00:00:00     
(11/12): ubi-7-rhscl/x86_64/updateinfo                             |   92 B  00:00:00     
(12/12): ubi-7-rhscl/x86_64/primary_db                             | 158 kB  00:00:00     
No package fuse-overlayfs available.
```

rhatdan: Thanks I could not pull that image myself.  Looks like I will need to wait for the upgrade to add fuseoverlayfs to that Dockerfile.

Dockerfile-dev
rhatdan: Done

rhatdan:
@nalind @adambkaplan PTAL

Dockerfile
```diff
@@ -5,17 +5,20 @@ RUN hack/build.sh
 
 FROM registry.svc.ci.openshift.org/openshift/origin-v4.0:base
 RUN INSTALL_PKGS=" \
-      which tar wget hostname sysvinit-tools util-linux \
-      socat tree findutils lsof bind-utils \
-      git tar bsdtar \
+      bsdtar findutils fuse-overlayfs git hostname lsof socat sysvinit-tools \
+      tar tree util-linux wget which \
```
nalind: Did we lose `bind-utils`?

Dockerfile-dev
```diff
@@ -4,17 +4,20 @@ LABEL io.k8s.display-name="OpenShift Origin Builder" \
       io.openshift.tags="openshift,builder"
 
 RUN INSTALL_PKGS=" \
-      which tar wget hostname sysvinit-tools util-linux \
-      socat tree findutils lsof bind-utils \
-      git tar bsdtar \
+      bsdtar findutils fuse-overlayfs git hostname lsof socat sysvinit-tools \
+      tar tree util-linux wget which \
```
nalind: Did we lose `bind-utils`?

rhatdan: Not intentionally

Dockerfile
rhatdan: Not intentionally. Fixed

adambkaplan:
@rhatdan build tests are currently broken due to an issue with cri-o. We're disabling the affected tests, hopefully that lands soon and we can re-test.

rhatdan:
@adambkaplan Should I rebase and repush?

adambkaplan:
@rhatdan - yes, underlying cri-o issue has been fixed and build tests are green.

Dockerfile.rhel7
```diff
@@ -5,17 +5,20 @@ RUN hack/build.sh
...
     yum clean all
 COPY --from=builder /go/src/github.com/openshift/builder/openshift-builder /usr/bin/
 COPY imagecontent/policy.json /etc/containers/
 COPY imagecontent/registries.conf /etc/containers/
 COPY imagecontent/storage.conf /etc/containers/
-RUN mkdir /var/cache/blobs
+RUN mkdir -p /var/cache/blobs \
+    /var/lib/shared/overlay-images \
+    /var/lib/shared/overlay-layers; \
+    touch /var/lib/shared/overlay-images/images.lock; \
```
adambkaplan: nit
```suggestion
    touch /var/lib/shared/overlay-images/images.lock && \
```

Dockerfile.rhel7
```diff
@@ -5,17 +5,20 @@ RUN hack/build.sh
...
     yum install -y $INSTALL_PKGS && \
     yum clean all
 COPY --from=builder /go/src/github.com/openshift/builder/openshift-builder /usr/bin/
 COPY imagecontent/policy.json /etc/containers/
 COPY imagecontent/registries.conf /etc/containers/
 COPY imagecontent/storage.conf /etc/containers/
-RUN mkdir /var/cache/blobs
+RUN mkdir -p /var/cache/blobs \
+    /var/lib/shared/overlay-images \
+    /var/lib/shared/overlay-layers; \
```
adambkaplan: nit
```suggestion
    /var/lib/shared/overlay-layers && \
```

Dockerfile
```diff
@@ -5,17 +5,20 @@ RUN hack/build.sh
...
     yum install -y $INSTALL_PKGS && \
     yum clean all
 COPY --from=builder /go/src/github.com/openshift/builder/openshift-builder /usr/bin/
 COPY imagecontent/policy.json /etc/containers/
 COPY imagecontent/registries.conf /etc/containers/
 COPY imagecontent/storage.conf /etc/containers/
-RUN mkdir /var/cache/blobs
+RUN mkdir -p /var/cache/blobs \
+    /var/lib/shared/overlay-images \
+    /var/lib/shared/overlay-layers; \
```
adambkaplan: nit
```suggestion
    /var/lib/shared/overlay-layers && \
```

Dockerfile
```diff
@@ -5,17 +5,20 @@ RUN hack/build.sh
...
+    touch /var/lib/shared/overlay-images/images.lock; \
```
adambkaplan: nit
```suggestion
    touch /var/lib/shared/overlay-images/images.lock && \
```

rhatdan:
@adambkaplan can we finally get this merged?

PR #80, opened 2019-06-24
Start using Buildah versus Docker for building container images

rhatdan:
Signed-off-by: Daniel J Walsh <dwalsh@redhat.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,7 @@
 IMAGE ?= docker.io/openshift/origin-docker-builder
 TAG ?= latest
 PROG  := openshift-builder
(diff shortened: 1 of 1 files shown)


rhatdan:
@nalind @bparees PTAL

rhatdan:
Well buildah is not planned to be run on a MAC, but Podman is, should I change this to podman run?

adambkaplan:
@rhatdan podman as default works, too (also helps with inner-loop development, since you can go directly from `podman build` to `podman run`)

bparees:
@adambkaplan it's your repo at this point so i'll let you sign off on this, but it looks ok to me.

/assign @adambkaplan

PR #87, opened 2019-07-31
clean up of some CA mount logic that is now handled by the build controller

gabemontero:
@openshift/openshift-team-developer-experience @bparees @nalind  FYI

/assign @adambkaplan 

An initial stab into modfications to openshift/builder in light of the 4.2 global CA injection work, along with reconciling some older TODOs I think are now handled by the build controller

My first openshift/builder PR :-)

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -46,25 +46,6 @@ func main() {
 		os.Exit(1)
 	}
 
-	// TODO: Remove this once the config-map based mount approach lands after rebase
-	oldServiceCASrc := fmt.Sprintf("%s/service-ca.crt", builder.SecretCertsMountPath)
-	oldServiceCADst := fmt.Sprintf("%s/service.crt", tlsCertRoot)
-	err = CopyFileIfExists(oldServiceCASrc, oldServiceCADst)
-	if err != nil {
-		fmt.Printf("Error setting up service CA cert: %v", err)
-		os.Exit(1)
-	}
-
-	// TODO: Remove this once the build controller mounts the internal registry's CA
-	// in certs.d/<internal-registry-hostname>/
-	newServiceCASrc := fmt.Sprintf("%s/service-ca.crt", builder.ConfigMapCertsMountPath)
-	newServiceCADst := fmt.Sprintf("%s/openshift-service.crt", tlsCertRoot)
-	err = CopyFileIfExists(newServiceCASrc, newServiceCADst)
-	if err != nil {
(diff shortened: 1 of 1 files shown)


bparees:
I don't think we should need changes here.  If things work as I *think* they will work, the build controller will be mounting the global CA content into /etc/pki.  That means it will be used by default, by everything.  (and if it's not sufficient the additional CAs that are injected for specific registries would be used)

gabemontero:
@bparees aside from the global proxy related removal what about the existing to-do's about removing code based on when existing function dropped, which it appears to have based on the code link provide?  Not worth potentially cleaning that stuff up?

bparees:
> @bparees aside from the global proxy related removal what about the existing to-do's about removing code based on when existing function dropped, which it appears to have based on the code link provide? Not worth potentially cleaning that stuff up?

i'd have to defer to @adambkaplan about whether it's safe to remove that stuff at this point or not(I haven't kept up with which certs are being mounted where in the builder pod these days and why), but if it is, then sure removing it makes sense, orthogonally.

gabemontero:
updates performed @adambkaplan .... PTAL

cmd/main.go
```diff
@@ -46,32 +46,6 @@ func main() {
...
-	// in certs.d/<internal-registry-hostname>/
-	newServiceCASrc := fmt.Sprintf("%s/service-ca.crt", builder.ConfigMapCertsMountPath)
-	newServiceCADst := fmt.Sprintf("%s/openshift-service.crt", tlsCertRoot)
-	err = CopyFileIfExists(newServiceCASrc, newServiceCADst)
-	if err != nil {
-		fmt.Printf("Error setting up service CA cert: %v", err)
-		os.Exit(1)
-	}
-
-	runtimeCASrc := fmt.Sprintf("%s/certs.d", builder.ConfigMapCertsMountPath)
```
adambkaplan: aaah! Don't delete this - this is the where the registry certs are mounted so they are trusted by the Docker client.

adambkaplan: (and buildah)

gabemontero:
ah ... thanks @adambkaplan .... commit updated ... PTAL again when able

gabemontero:
terraform flake on e2e-aws

/test e2e-aws

gabemontero:
aws constraint flake on image-eco

/test e2e-aws-image-ecosystem

gabemontero:
bump @adambkaplan ... tests are passing, let's merge this one

PR #88, opened 2019-08-07
Remove fuse-overlayfs from Dockerfiles

adambkaplan:
fuse-overlayfs is not available in RHEL 7.
This is causing ART builds for 4.2 to fail.


rhatdan:
LGTM

PR #89, opened 2019-08-20
Bug 1716697: Strip the tag from the reference we print after pushing an image

nalind:
When we combine the image name that we pushed to with a digest in order to print the "Successfully pushed..." message, strip out the tag.  Will need to be cherry-picked to fix https://bugzilla.redhat.com/show_bug.cgi?id=1716697.


adambkaplan:
/retitle Bug 1716697: Strip the tag from the reference we print after pushing an image

adambkaplan:
@nalind we are still on 4.2 for this repo. The bug is not severe enough that I would consider backporting to 4.1.

PR #90, opened 2019-08-20
[release-4.2] Bug 1716697: strip the tag from the reference we print after pushing

nalind:
When we combine the image name that we pushed to with a digest in order to print the "Successfully pushed..." message, strip out the tag.  Cherry-picked from #89.


adambkaplan:
@nalind please open this PR against master. We are not planning on opening master to 4.3 work yet.

nalind:
@adambkaplan #89 is the same patch as this PR, but for the master branch.

adambkaplan:
/close

Duplicates #89

PR #91, opened 2019-08-20
[release-4.3] strip the tag from the reference we print after pushing

nalind:
When we combine the image name that we pushed to with a digest in order to print the "Successfully pushed..." message, strip out the tag.  Cherry-picked from #89.


adambkaplan:
/close

Duplicates #89

PR #92, opened 2019-08-29
Bug 1746499: Add ":latest" to push tag if none is specified

nalind:
Use `util.AddImageNames()` to add the `pushTag` to our just-built image, so that the name we apply will include a "latest" tag if one wasn't directly specified.

The `AddImageNames()` function ultimately expands the name we give it, in the same way that Push() does with the name we give it, so this should resolve cases where we would add a name with `SetNames()` that `Push()` would then be unable to use to find the image.


adambkaplan:
/retitle Bug 1746499: Add ":latest" to push tag if none is specified

PR #93, opened 2019-08-30
Bump github.com/containers/{storage,image,buildah}

nalind:
Bump github.com/containers/storage from v1.12.13 to v1.13.2, github.com/containers/image from v2.0.0 to v3.0.2 (and its github.com/vbauerster/mpb requirement from v3.3.4 to v3.4.0), github.com/containers/buildah from v1.9.1 to v1.10.1, and pin k8s.io/apimachinery, k8s.io/cli-runtime, and openshift/api to the versions used by OpenShift 4.2.


glide.yaml
```diff
@@ -4,17 +4,21 @@ import:
...
 - package: k8s.io/kubernetes
   repo:    https://github.com/openshift/kubernetes.git
   version: origin-4.2-kubernetes-1.14.0
+- package: k8s.io/cli-runtime
+  repo:    https://github.com/openshift/kubernetes-cli-runtime.git
+  version: origin-4.2-kubernetes-1.14.0
 # 2. openshift
 - package: github.com/openshift/api
-  version: master
+  version: release-4.2
```
adambkaplan: If you're switching api to `release-4.2`, switch `client-go` as well.

glide.yaml
```diff
@@ -4,17 +4,21 @@ import:
 - package: k8s.io/api
   version: kubernetes-1.14.0
 - package: k8s.io/apimachinery
-  version: kubernetes-1.14.0
+  repo:    https://github.com/openshift/kubernetes-apimachinery.git
+  version: origin-4.2-kubernetes-1.14.0
```
adambkaplan: Is there a reason why we need to use our fork? Is there a carry patch there that we need?

glide.yaml
nalind: Ah, didn't check for that.  Changing it.

glide.yaml
nalind: We started needing `k8s.io/apimachinery/pkg/apis/meta/v1beta1.AddToScheme()` in `k8s.io/kubernetes/pkg/kubectl/scheme.init()` (in https://github.com/openshift/kubernetes.git's origin-4.2-kubernetes-1.14.0 branch).

glide.yaml
nalind: Ah, `openshift/client-go` isn't `k8s.io/client-go`.  Pinning it.

adambkaplan:
/retitle Bug 1754695: Bump containers/{storage,image,buildah}

PR #95, opened 2019-09-09
Bug 1747937: Forwarding HTTPProxy environment variables.

ricardomaraschini:
This patch sets HTTPProxy option on buildah.CommonBuildOptions to true,
that will cause Buildah to forward HTTPProxy environment variables down
the line.


gabemontero:
ahh .... we missed this in our openshift/builder exploration for the build/proxy card @adambkaplan :-/

PR #96, opened 2019-09-09
Bug 1745789: toggle ssh parameters if ssh-auth source secret includes a known_host…

gabemontero:
…s setting

@openshift/openshift-team-developer-experience @nalind FYI

/assign @adambkaplan 

FYI ... developing a unit test became untenable because
- either `make test` would just hang on me ... even without my commit
- or if I ran `go test ..` directly, I got:

```
# pkg-config --cflags  -- devmapper
Package devmapper was not found in the pkg-config search path.
Perhaps you should add the directory containing `devmapper.pc'
to the PKG_CONFIG_PATH environment variable
Package 'devmapper', required by 'virtual:world', not found
pkg-config: exit status 1
# github.com/openshift/builder/vendor/github.com/containers/storage/drivers/btrfs
vendor/github.com/containers/storage/drivers/btrfs/btrfs.go:8:10: fatal error: btrfs/ioctl.h: No such file or directory
 #include <btrfs/ioctl.h>
          ^~~~~~~~~~~~~~~
compilation terminated.
FAIL	github.com/openshift/builder/pkg/build/builder/cmd/scmauth [build failed]
```

If openshift/builder veterans have pointers, please provide


nalind:
We turn off various features of containers/storage and containers/image which we either don't intend to use, or which aren't going to be used from inside of a container.  If you `source` `hack/lib/constants.sh` and then add a `-tags $OS_GOFLAGS_TAGS_TEST` to your `go test` invocation, you'll run tests using the same build tags that `make test` and `hack/test.sh` use, which should help.

pkg/build/builder/cmd/scmauth/sshkey.go
```diff
@@ -21,10 +23,33 @@ func (_ SSHPrivateKey) Setup(baseDir string, context SCMAuthContext) error {
...
+			foundPrivateKey = true
+		}
+		log.V(5).Infof("source secret dir %s has file %s", baseDir, file.Name())
+	}
+	if !foundPrivateKey {
+		return fmt.Errorf("could not find the ssh-privatekey file for the ssh secret stored at %s", baseDir)
+	}
+	// let's see if known_hosts was included in the secret
+	knownHostsPath := filepath.Join(baseDir, kownHostsFileName)
 	content := "#!/bin/sh\nssh -i " +
```
ricardomaraschini: I have the impression that now this `-i` is lost as it does not contain, as flag value, the path for the private key. The command become something like:

```
ssh -i -o StrictHostKeyChecking=false $@
```

gabemontero: yep I inadvertently dropped the filename ... will address - thanks

pkg/build/builder/cmd/scmauth/sshkey.go
```diff
@@ -1,11 +1,13 @@
 package scmauth
 
 import (
+	"fmt"
 	"io/ioutil"
 	"path/filepath"
 )
 
 const SSHPrivateKeyMethodName = "ssh-privatekey"
+const kownHostsFileName = "known_hosts"
```
adambkaplan: nit:
```suggestion
const knownHostsFileName = "known_hosts"
```

gabemontero:
Agree on the unit tests @adambkaplan , I just can't get them to run locally, even after @nalind 's suggestions

At worst, I might start the 3.11 version of this and develop the unit test there, without all the containers/image and storage noise that is apparently making running them out of openshift/builder problematic (at least in the short term, given our runway  this week), and then just copy them to the 4.2 version.

I'll be taking another stab at the 4.2 unit tests "natively" first this afternoon before doing that.

pkg/build/builder/cmd/scmauth/sshkey.go
gabemontero: fixed / thanks

gabemontero:
ok sorted out the `make test` issue @adambkaplan (a combination of a really long initial compile which @nalind also observed with my .bashrc GOPATH still overriding the GOPATH I set manually before running `make test`

PTAL / bump

gabemontero:
OK we have clean tests @adambkaplan for this CVE fix

pkg/build/builder/cmd/scmauth/sshkey.go
```diff
@@ -21,10 +23,32 @@ func (_ SSHPrivateKey) Setup(baseDir string, context SCMAuthContext) error {
...
+		case file.Name() == SSHPrivateKeyMethodName:
+			foundPrivateKey = true
+		}
+		log.V(5).Infof("source secret dir %s has file %s", baseDir, file.Name())
+	}
+	if !foundPrivateKey {
+		return fmt.Errorf("could not find the ssh-privatekey file for the ssh secret stored at %s", baseDir)
+	}
+	// let's see if known_hosts was included in the secret
+	knownHostsPath := filepath.Join(baseDir, knownHostsFileName)
```
bparees: this could be moved down into the else block since that's the only path where it's used.

gabemontero: update to move down pushed thanks

gabemontero:
flakes circumnavigated @bparees and @adambkaplan ... clean tests and current comments resolved
aside from @adambkaplan 's query about additional reviewers

PTAL / bump

bparees:
it lgtm, might be worth an e2e test that provides a secret w/ the github knownhost entry to truly validate e2e (and one that provides an invalid knownhost entry and confirm it fails).

bparees:
(since that test probably has to be contributed to origin, you probably have to merge this before you can land such a test)

gabemontero:
> it lgtm, might be worth an e2e test that provides a secret w/ the github knownhost entry to truly validate e2e (and one that provides an invalid knownhost entry and confirm it fails).

we would also have to provide a valid ssh private key in creating that secret; that is still required with this PR .... seems problematic to provide one of our private keys, or is there an "anonymous" private key that could be provided that I should be aware of?

inventory note: there are not git ssh extended tests in general that I could find at this time;  there is a username/password one using the old internal git server that is currently being skipped:  https://github.com/openshift/origin/blob/master/test/extended/builds/gitauth.go#L77

bparees:
> we would also have to provide a valid ssh private key in creating that secret; that is still required with this PR

hmm, good point.  i can't think of a good solution offhand that doesn't involve running our own git server (which as you note, we used to do for some tests but was problematic to maintain...  we probably should have a tech-debt item to get back to a working state for that though, so we can test auth scenarios in an e2e again)

PR #99, opened 2019-09-16
WIP: Bug 1750650: switch git install to RH SCL to get later version supporting CA imports

gabemontero:
… CA import for free

/assign @bparees 
/assign @ricardomaraschini 
/assign @adambkaplan 

@openshift/openshift-team-developer-experience PTAL

with @thiagoalessio 's https://gitlab.cee.redhat.com/openshift-art/ocp-build-data/commit/077eb2ab331c5aadadd5bf6c22644533c526ae43 enabled + this PR we can obtain 2.3-5.el7

which allows us to not have to specify sslCAPath in the gitconfig file to pick up the global CA mouted cert.

obsoletes https://github.com/openshift/builder/pull/98

will mark this with https://bugzilla.redhat.com/show_bug.cgi?id=1750650 after
- ideally QE validates some of our findings on their own though if need be we won't block on this
- we have some passing tests
- we get consensus on the change

PR #100, opened 2019-09-16
ensure global CA injected certs are included in gitconfig sslCAInfo

gabemontero:
/assign @bparees 
/assign @adambkaplan 
/assign @ricardomaraschini 

this is the alternative to https://github.com/openshift/builder/pull/99 where we don't bump git to the later version provided by SCL on rhel7, but instead always insure that the cert injected by global CA support is included in the gitconfig

if we go with this approach instead I'll tag this bug to https://bugzilla.redhat.com/show_bug.cgi?id=1750650

### DIFF
--- a/pkg/build/builder/cmd/builder.go
+++ b/pkg/build/builder/cmd/builder.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"strings"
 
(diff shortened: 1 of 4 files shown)


gabemontero:
Upon further reivew, it has been determined that updating sslCAInfo was not sufficient to solve the https proxy situation 

We need to update the curl libs, either directly, or via upgrading git

PR #101, opened 2019-09-18
[4.3] Bug 1753247: listen for sigterm and terminate

rphillips:
kubernetes will send a sigterm to signal for a graceful shutdown. builder should listen for the sigterm and exit cleanly without requiring a sigkill.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -6,8 +6,10 @@ import (
 	"io/ioutil"
 	"math/rand"
 	"os"
+	"os/signal"
 	"path/filepath"
 	"runtime"
+	"syscall"
 	"time"
 
 	"github.com/containers/storage/pkg/reexec"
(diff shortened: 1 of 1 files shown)


adambkaplan:
/lgtm cancel

Need to think about which exit code we should send - SIGTERM means the build didn't complete and should fail.

rphillips:
@adambkaplan I changed the return code to 1, since, it appears, the rest of the code will also return `1` on error. Just let me know if we should change it again!

cmd/main.go
```diff
@@ -25,6 +27,13 @@ func main() {
 		return
 	}
 
+	sigs := make(chan os.Signal, 1)
+	signal.Notify(sigs, syscall.SIGTERM)
+	go func() {
+		<-sigs
+		os.Exit(1)
```
adambkaplan: nit - print a message indicating we received an unexpected terminate signal

```suggestion
		fmt.Println("Error: received unexpected terminate signal")
		os.Exit(1)
```

rphillips: Updated. Thanks!

rphillips:
4.3 is open 

/hold cancel

rphillips:
@adambkaplan friendly ping for a review and approval

adambkaplan:
/hold

Got bit by the `oc -p` deprecation

rphillips:
@adambkaplan Do I need to update the PR?

gabemontero:
the `oc exec -p` test fix has merged

/hold cancel

/retest

PR #102, opened 2019-09-24
Bug 1745192: seed containers/image with entire dockerconfig for authentication in …

gabemontero:
…case mirrors are in play

a possible solution for https://bugzilla.redhat.com/show_bug.cgi?id=1745192 though we should update openshift/source-to-image and bump back here to provide the complete dockerconfig 

/assign @adambkaplan 
/assign @nalind 
/assign @bparees
@openshift/openshift-team-developer-experience 

the other solution that comes to my mind is sucking in the registry.conf, determine if there is a mirror for the image we have,  and grabs its creds instead .... seems too smart for its own good but I throw that out there


pkg/build/builder/sti.go
```diff
@@ -226,12 +228,14 @@ func (s *S2IBuilder) Build() error {
 
 	// If DockerCfgPath is provided in buildapiv1.Config, then attempt to read the
 	// dockercfg file and get the authentication for pulling the images.
+	//TODO in theory change to GetAllDockerAuths once PullAuthentication can take multiples
 	t, _ := dockercfg.NewHelper().GetDockerAuth(config.BuilderImage, dockercfg.PullAuthType)
 	config.PullAuthentication = s2iapi.AuthConfig{Username: t.Username, Password: t.Password, Email: t.Email, ServerAddress: t.ServerAddress}
```
gabemontero: so yeah, this field from the openshift/source-to-image repo only takes a single registry auth at the moment

gabemontero: the `result, err := builder.Build(config)` call further down in this file ultimately consumes it

gabemontero:
In the last image_eco got an `Unauthorized` during `hot deploy for openshift ruby image Rails example should work with hot deploy` and then in `hot deploy for openshift perl image hot deploy test should work` it seemed the modify step was just "slow" and then the test gave up:

```
STEP: modifying the source code with enabled hot deploy
Sep 25 13:03:30.598: INFO: Creating new exec pod
Sep 25 13:03:41.700: INFO: Waiting up to 3m0s to wget http://10.128.2.29:8080/
Sep 25 13:03:41.700: INFO: Running '/usr/bin/kubectl --server=https://api.ci-op-br61v5wm-28622.origin-ci-int-aws.dev.rhcloud.com:6443 --kubeconfig=/tmp/admin.kubeconfig exec --namespace=e2e-test-s2i-perl-r4srj 2ac57b4a-0b41-4d5e-ba8b-c7563d84a3cc -- /bin/sh -x -c curl -vvv http://10.128.2.29:8080/'
Sep 25 13:03:42.313: INFO: stderr: "+ curl -vvv http://10.128.2.29:8080/\n*   Trying 10.128.2.29...\n* TCP_NODELAY set\n  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current\n                                 Dload  Upload   Total   Spent    Left  Speed\n\r  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0* Connected to 10.128.2.29 (10.128.2.29) port 8080 (#0)\n> GET / HTTP/1.1\r\n> Host: 10.128.2.29:8080\r\n> User-Agent: curl/7.61.1\r\n> Accept: */*\r\n> \r\n< HTTP/1.1 200 OK\r\n< Date: Wed, 25 Sep 2019 13:03:42 GMT\r\n< Server: Apache/2.4.34 (Red Hat) mod_perl/2.0.9 Perl/v5.24.0\r\n< Transfer-Encoding: chunked\r\n< Content-Type: text/plain; charset=UTF-8\r\n< \r\n{ [24 bytes data]\n\r100    14    0    14    0     0    933      0 --:--:-- --:--:-- --:--:--   933\n* Connection #0 to host 10.128.2.29 left intact\n"
Sep 25 13:03:42.313: INFO: stdout: "initial value\n"
Sep 25 13:03:43.408: INFO: Running 'oc --namespace=e2e-test-s2i-perl-r4srj --config=/tmp/configfile800998646 exec perl-2-65kcd -c perl -- sed -ie s/initial value/modified value/ lib/My/Test.pm'
RunInPodContainer exec output: 
Sep 25 13:03:44.908: INFO: Creating new exec pod
Sep 25 13:03:57.986: INFO: Waiting up to 3m0s to wget http://10.128.2.29:8080/
Sep 25 13:03:57.986: INFO: Running '/usr/bin/kubectl --server=https://api.ci-op-br61v5wm-28622.origin-ci-int-aws.dev.rhcloud.com:6443 --kubeconfig=/tmp/admin.kubeconfig exec --namespace=e2e-test-s2i-perl-r4srj 5504f85f-2a3e-4742-b922-d7b0b947318d -- /bin/sh -x -c curl -vvv http://10.128.2.29:8080/'
Sep 25 13:03:58.466: INFO: stderr: "+ curl -vvv http://10.128.2.29:8080/\n*   Trying 10.128.2.29...\n* TCP_NODELAY set\n  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current\n                                 Dload  Upload   Total   Spent    Left  Speed\n\r  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0* Connected to 10.128.2.29 (10.128.2.29) port 8080 (#0)\n> GET / HTTP/1.1\r\n> Host: 10.128.2.29:8080\r\n> User-Agent: curl/7.61.1\r\n> Accept: */*\r\n> \r\n< HTTP/1.1 200 OK\r\n< Date: Wed, 25 Sep 2019 13:03:58 GMT\r\n< Server: Apache/2.4.34 (Red Hat) mod_perl/2.0.9 Perl/v5.24.0\r\n< Transfer-Encoding: chunked\r\n< Content-Type: text/plain; charset=UTF-8\r\n< \r\n{ [25 bytes data]\n\r100    15    0    15    0     0   3000      0 --:--:-- --:--:-- --:--:--  3000\n* Connection #0 to host 10.128.2.29 left intact\n"
Sep 25 13:03:58.466: INFO: stdout: "modified value\n"
[AfterEach] 
  /go/src/github.com/openshift/origin/_output/local/go/src/github.com/openshift/origin/test/extended/image_ecosystem/s2i_perl.go:38
```

pkg/build/builder/sti.go
gabemontero: or wait, do we even care here ? .... s2i is suppose to be the asDockerfile in 4.x .... investigating

gabemontero: Yeah the auths are gathered from the `pullAuthConfigs, err := s.setupPullSecret()` call further down in this method

And the dockerfile builder in s2i does not make any RPCs over the docker client

I'd contend that these 2 lines can be deleted.  

I may do that in a separate commit and see what the CI results show us.

adambkaplan:
Perhaps in a separate PR we can discuss why we are using `k8s.io/pkg/credentialprovider`. This is kubelet code that we probably should get rid of.

gabemontero:
> Perhaps in a separate PR we can discuss why we are using `k8s.io/pkg/credentialprovider`. This is kubelet code that we probably should get rid of.

Yeah it was a coin flip for me wrt k8s structs vs. the fsouza ones.  Both seem to have pros/cons, as does crafting something new.

gabemontero:
First set of clean test runs.

I still want to try this with removal of that PullAuthentication setting I mentioned above.

gabemontero:
ok I've pushed the commit that removes the config.PullAuthentication initialization

gabemontero:
yep the tests pass  .... officially deleting / squashing

gabemontero:
OK I've taken off the WIP designation @openshift/openshift-team-developer-experience 

@adambkaplan @bparees @mtrmac PTAL

pkg/build/builder/cmd/dockercfg/cfg.go
```diff
@@ -40,6 +40,28 @@ func NewHelper() *Helper {
 func (h *Helper) InstallFlags(flags *pflag.FlagSet) {
 }
 
+// GetAllDockerAuths returns the entire docker config found, regardless of filtering
+// by imageName, and whether it was readfrom a file
+func (h *Helper) GetAllDockerAuths(authType string) (credentialprovider.DockerConfig, bool) {
```
adambkaplan: Why not return `error` instead of `bool`?

pkg/build/builder/docker.go
```diff
@@ -147,12 +150,20 @@ func (d *DockerBuilder) Build() error {
...
 		)
+		ref, err := reference.Parse(pushTag)
+		if err != nil {
+			log.V(0).Infof("error parsing image name %s: %v", ref, err)
+		}
+		if ref.Registry == "" {
+			log.V(2).Infof("defaulting registry to docker.io for image %s", pushTag)
+			ref.Registry = "docker.io"
+		}
+		auth, _ := pushAuthConfig[ref.Registry]
```
adambkaplan: I think this should be:
```suggestion
		auth, authPresent := pushAuthConfig[ref.Registry]
```

pkg/build/builder/docker.go
```diff
@@ -147,12 +150,20 @@ func (d *DockerBuilder) Build() error {
 
 	if push && pushTag != "" {
 		// Get the Docker push authentication
-		pushAuthConfig, authPresent := dockercfg.NewHelper().GetDockerAuth(
-			pushTag,
+		pushAuthConfig, authPresent := dockercfg.NewHelper().GetAllDockerAuths(
```
adambkaplan: Probably don't need `authPresent` here if below you check if the push config exists for the _particular_ registry.

pkg/build/builder/sti.go
```diff
@@ -226,12 +228,11 @@ func (s *S2IBuilder) Build() error {
 
 	// If DockerCfgPath is provided in buildapiv1.Config, then attempt to read the
 	// dockercfg file and get the authentication for pulling the images.
-	t, _ := dockercfg.NewHelper().GetDockerAuth(config.BuilderImage, dockercfg.PullAuthType)
-	config.PullAuthentication = s2iapi.AuthConfig{Username: t.Username, Password: t.Password, Email: t.Email, ServerAddress: t.ServerAddress}
 
 	if s.build.Spec.Strategy.SourceStrategy.ForcePull || !isImagePresent(s.dockerClient, config.BuilderImage) {
 		startTime := metav1.Now()
-		err = s.pullImage(config.BuilderImage, t)
+		auth, _ := dockercfg.NewHelper().GetAllDockerAuths(dockercfg.PullAuthType)
```
adambkaplan: returning `error` instead of `bool` is instructive here. In this case we probably would not want to return error if the auth config could not be read. We'd fall back to pulling an image anonymously.

pkg/build/builder/cmd/dockercfg/cfg.go
gabemontero: I followed the existing pattern of `GetDockerAuth` which was the method that used to be employed that return only the auth for the registry in question

But I don't feel strongly about `bool` vs. `error` if you want me to change that aspect as well from the existing implementation.

So, "change" or "no change" ?

pkg/build/builder/docker.go
gabemontero: agreed

pkg/build/builder/docker.go
gabemontero: agreed .... and of course, whether you have auth for the actual registry is not as critical in the mirror scenario, as long as you have the auth for the mirror

pkg/build/builder/sti.go
gabemontero: in the current implementation we only return false when an error occurred so they are equivalent in that sense ... we no longer care which auths are present, and are in fact still trying to pull anonymously

in the other file, the use of authPresent ultimately was only used for triggering some debug

gabemontero:
pushed updates for 2 of your comments @adambkaplan 

discussion open for the other 2

pkg/build/builder/sti.go
```diff
@@ -226,12 +228,11 @@ func (s *S2IBuilder) Build() error {
 
 	// If DockerCfgPath is provided in buildapiv1.Config, then attempt to read the
 	// dockercfg file and get the authentication for pulling the images.
-	t, _ := dockercfg.NewHelper().GetDockerAuth(config.BuilderImage, dockercfg.PullAuthType)
-	config.PullAuthentication = s2iapi.AuthConfig{Username: t.Username, Password: t.Password, Email: t.Email, ServerAddress: t.ServerAddress}
```
mtrmac: Delete the comment as well?

(Also noting that the `PullAuthentication` field was not removed, and some code in source-to-image still refers to it. From a very quick look it seems that it can’t be used (it would crash on a `nil` interface submitted by `runtimeBuilderFactory.Builder`), but I may well be not looking hard enough.)

pkg/build/builder/daemonless.go
```diff
@@ -272,11 +274,13 @@ func pushDaemonlessImage(sc types.SystemContext, store storage.Store, imageName
...
 
-	if authConfig.Username != "" && authConfig.Password != "" {
-		log.V(2).Infof("Setting authentication secret for %q.", authConfig.ServerAddress)
-		systemContext.DockerAuthConfig = &types.DockerAuthConfig{
-			Username: authConfig.Username,
-			Password: authConfig.Password,
+	if len(authConfig) != 0 {
+		for reg, auth := range authConfig {
+			log.V(2).Infof("Setting authentication secret for %q.", reg)
+			systemContext.DockerAuthConfig = &types.DockerAuthConfig{
```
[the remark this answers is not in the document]
mtrmac: (Pushing currently doesn’t push to mirrors anyway, so submitting only one credential via `DockerAuthConfig` is not in itself actually wrong, but it would have to be the credential for the primary registry, not for any of the mirrors. Better to just provide all of them anyway.)

pkg/build/builder/cmd/dockercfg/cfg.go
```diff
@@ -40,6 +40,28 @@ func NewHelper() *Helper {
...
+	log.V(3).Infof("Getting docker config in paths : %v", searchPaths)
+	cfg, err = GetDockerConfig(searchPaths)
+	if err != nil {
+		klog.Errorf("Reading docker config from %v failed: %v", searchPaths, err)
+		return credentialprovider.DockerConfig{}, false
+	}
+	return cfg, true
+}
+
 // GetDockerAuth returns a valid Docker AuthConfiguration entry, and whether it was read
```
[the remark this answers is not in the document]
gabemontero: Correct it is no longer called.

I left it on on purpose in case it was needed in the future.  The use of the keyring stuff was foreign enough to me that I wanted it around for reference.

pkg/build/builder/daemonless.go
```diff
@@ -25,7 +27,7 @@ import (
 
 	buildapiv1 "github.com/openshift/api/build/v1"
 	builderutil "github.com/openshift/builder/pkg/build/builder/util"
-	"github.com/openshift/library-go/pkg/image/reference"
+	//"github.com/openshift/library-go/pkg/image/reference"
```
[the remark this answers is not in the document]
gabemontero: yep meant to do that and forgot - thanks

pkg/build/builder/sti.go
gabemontero: I did an detailed examination, and the "as dockerfile" path in source-to-image does not user PullAuthentication.

I believe the comment still has relevance for the subsequent code below.

pkg/build/builder/cmd/dockercfg/cfg.go
```diff
@@ -125,9 +147,9 @@ func ReadDockerConfigJsonFileGeneratedFromSecret(path []string) (cfg credentialp
 	return nil, err
 }
 
-//getExtraSearchPaths get extra paths that may contain docker-config type files.
+// GetExtraSearchPaths get extra paths that may contain docker-config type files.
 //this invocation we do not need to handle user.Current() since upstream k8s have handled HOME path
-func getExtraSearchPaths() (searchPaths []string) {
+func GetExtraSearchPaths() (searchPaths []string) {
```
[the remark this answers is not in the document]
gabemontero: hmmm ... perhaps an artifact from an prior incarnation of this change I ultimately did not push up.

Will double check and adjust accordingly, thanks.

gabemontero:
/hold

based on @mtrmac 's https://github.com/openshift/builder/pull/102#discussion_r328896783

pkg/build/builder/cmd/dockercfg/cfg.go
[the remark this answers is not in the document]
mtrmac: Works for me.

pkg/build/builder/cmd/dockercfg/cfg.go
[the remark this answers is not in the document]
gabemontero: corrected in next push

pkg/build/builder/daemonless.go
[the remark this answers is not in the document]
gabemontero: corrected in next push

pkg/build/builder/daemonless.go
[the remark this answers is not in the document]
gabemontero: rather than even specifying the temp file @mtrmac , perhaps we should just set `systemContext.AuthFilePath` to the location from the search path currently determined in `GetAllDockerAuths` ?

Something akin to the code above that was previously commented out.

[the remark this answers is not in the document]
mtrmac: Yes, sure. I somehow didn’t connect the reading and writing code.

(Do note the caveat about letting the creator of that file (= anyone who can edit the secret object, it seems) specify arbitrary credential helpers.)

pkg/build/builder/sti.go
gabemontero: with the responses to @mtrmac comments I believe this has become a no-op

pkg/build/builder/cmd/dockercfg/cfg.go
gabemontero: again with the changes stemming from @mtrmac 's comments this has become a no-op

pkg/build/builder/sti.go
gabemontero: again no-op with new changes

pkg/build/builder/daemonless.go
```diff
@@ -63,18 +65,18 @@ func pullDaemonlessImage(sc types.SystemContext, store storage.Store, imageName
...
 	if ref.Registry == "" {
 		log.V(2).Infof("defaulting registry to docker.io for image %s", imageName)
 		ref.Registry = "docker.io"
-	}
+	}*/
 
-	if authConfig.Username != "" && authConfig.Password != "" {
-		log.V(5).Infof("Setting authentication for registry %q for %q.", ref.Registry, imageName)
-		if err := config.SetAuthentication(&systemContext, ref.Registry, authConfig.Username, authConfig.Password); err != nil {
+	for reg, auth := range authConfig {
```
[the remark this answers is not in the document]
gabemontero: ok I just pushed the pivot on this

pkg/build/builder/daemonless.go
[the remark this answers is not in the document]
gabemontero: update on this new direction pushed

pkg/build/builder/docker.go
```diff
@@ -147,12 +150,20 @@ func (d *DockerBuilder) Build() error {
...
 		)
+		ref, err := reference.Parse(pushTag)
+		if err != nil {
+			log.V(0).Infof("error parsing image name %s: %v", ref, err)
+		}
+		if ref.Registry == "" {
+			log.V(2).Infof("defaulting registry to docker.io for image %s", pushTag)
+			ref.Registry = "docker.io"
+		}
+		auth, authPresent := pushAuthConfig[ref.Registry]
```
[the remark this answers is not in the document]
gabemontero: update to new direction pushed

gabemontero:
/hold cancel

ok I've refactored based on @mtrmac 's input from late last week.

based on the nature of the change, I squashed the commits

@mtrmac @adambkaplan PTAL

pkg/build/builder/daemonless_unsupported.go
```diff
@@ -24,7 +26,7 @@ type DaemonlessClient struct {
 func (d *DaemonlessClient) BuildImage(opts docker.BuildImageOptions) error {
 	return errors.New("building images not supported on this platform")
 }
-func (d *DaemonlessClient) PushImage(opts docker.PushImageOptions, auth docker.AuthConfiguration) error {
+func (d *DaemonlessClient) PushImage(opts docker.PushImageOptions, auth credentialprovider.DockerConfig) error {
```
gabemontero: oops missed this one

pkg/build/builder/daemonless.go
```diff
@@ -61,23 +61,7 @@ func pullDaemonlessImage(sc types.SystemContext, store storage.Store, imageName
 	// if credsDir, ok := os.LookupEnv("PULL_DOCKERCFG_PATH"); ok {
 	// 	systemContext.AuthFilePath = filepath.Join(credsDir, "config.json")
 	// }
-	systemContext.AuthFilePath = "/tmp/config.json"
```
mtrmac: (Maybe delete the comment above as well, in all of the modified functions.)

pkg/build/builder/docker.go
```diff
@@ -147,16 +146,18 @@ func (d *DockerBuilder) Build() error {
...
-			dockercfg.PushAuthType,
-		)
-		if authPresent {
-			log.V(4).Infof("Authenticating Docker push with user %q", pushAuthConfig.Username)
+		searchPaths := dockercfg.NewHelper().GetDockerAuthSearchPaths(dockercfg.PushAuthType)
+		ref, err := reference.Parse(pushTag)
+		if err != nil {
+			log.V(0).Infof("error parsing image name %s: %v", ref, err)
+		}
+		if ref.Registry == "" {
```
mtrmac: (The `reference.Parse`+`"docker.io"` default calls could be moved insider `reportPushFailure`, the only consumer.)

gabemontero:
e2e-aws failure was terraform flake

/test e2e-aws

gabemontero:
`oc exec` may have changed and hence is breaking image_ecosystem:

```
STEP: calling the binary using 'oc exec /bin/bash -c'
Sep 30 19:51:29.334: INFO: Running 'oc --namespace=e2e-test-s2i-usage-sqdw7 --config=/tmp/configfile059334108 exec -p test-pod-37f3db13-2598-4b01-b8d6-3ddf6b4b2aba -- /bin/bash -c ruby --version'
Sep 30 19:51:29.429: INFO: Error running &{/usr/bin/oc [oc --namespace=e2e-test-s2i-usage-sqdw7 --config=/tmp/configfile059334108 exec -p test-pod-37f3db13-2598-4b01-b8d6-3ddf6b4b2aba -- /bin/bash -c ruby --version] []   Error: unknown shorthand flag: 'p' in -p
```

asking around

pkg/build/builder/docker.go
gabemontero: getting some very odd unit test errors on this one ... gonna punt on it

pkg/build/builder/daemonless.go
gabemontero: part of next push

pkg/build/builder/cmd/dockercfg/cfg.go
```diff
@@ -40,6 +42,21 @@ func NewHelper() *Helper {
 func (h *Helper) InstallFlags(flags *pflag.FlagSet) {
 }
 
+// GetDockerAuthSearchPaths returns the list of possible locations for the raw
+// docker auth file that we'll inspect, where the winning location will be set
+// into the containers/image SystemContext.AuthFilePath field
+func (h *Helper) GetDockerAuthSearchPaths(authType string) []string {
```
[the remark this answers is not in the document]
gabemontero: part of next push

pkg/build/builder/cmd/dockercfg/cfg.go
```diff
@@ -103,6 +120,49 @@ func GetDockercfgFile(path string) string {
...
+		return false
+	}
+	return true
+}
+
+// GetDockerConfigPath returns the first path that provides a valid DockerConfig;
+// modified elements from credentialprovider.ReadDockerConfigJSONFile and
+// credentialprovider.ReadDockercfgFile to get at path vs. credentialprovider.DockerConfig
+func GetDockerConfigPath(paths []string) string {
+	if len(paths) == 0 {
```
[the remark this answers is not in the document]
gabemontero: part of next push

gabemontero:
thanks @mtrmac 

I believe we are good on the users who have edit secret permissions question, but @adambkaplan should chime in that, as well as generally review the rework here.

gabemontero:
I've confirmed that `oc exec` has changed.  I've opened https://bugzilla.redhat.com/show_bug.cgi?id=1757424 to track my updating of those test cases

gabemontero:
pulls and pushes are failing with in e2e-aws-builds

our new approach for passing in the docker authentication is not working @mtrmac and @adambkaplan 

I'll start debugging locally and advise when I make some discoveries.

gabemontero:
https://github.com/openshift/origin/pull/23891 is opened for the `oc exec` issues ... there was a hit in the extended builds suite as well, though I suspect that is getting skipped somehow (though I did not investigate to confirm).

gabemontero:
ok I've updated the commit @mtrmac @adambkaplan after seeing some success with local testing

Basically I had to add checks for `.dockercfg` and `.dockerconfigjson` in addition to `config.json`

pkg/build/builder/cmd/dockercfg/cfg.go
```diff
@@ -103,6 +116,48 @@ func GetDockercfgFile(path string) string {
...
+	return true
+}
+
+// GetDockerConfigPath returns the first path that provides a valid DockerConfig;
+// modified elements from credentialprovider methods called from GetDockerConfig, following the same order of precedenced
+// articulated in GetDockerConfig, via the order of file names set in dockerFilesToExamine
+func GetDockerConfigPath(paths []string) string {
+	for _, configPath := range paths {
+		for _, file := range dockerFilesToExamine {
+			absDockerConfigFileLocation, err := filepath.Abs(filepath.Join(configPath, file))
```
mtrmac: `GetDockerConfig` == `credentialprovider.ReadDockerConfigJSONFile` only tries `configJsonFileName = "config.json"`, not the other two, AFAICS, so I’m not sure why all three need to be checked here.  It doesn’t quite _hurt_, of course.

gabemontero: If you look at https://github.com/openshift/builder/blob/master/pkg/build/builder/cmd/dockercfg/cfg.go#L107-L114

```
func GetDockerConfig(path []string) (cfg credentialprovider.DockerConfig, err error) {
	if cfg, err = credentialprovider.ReadDockerConfigJSONFile(path); err != nil {
		if cfg, err = ReadDockerConfigJsonFileGeneratedFromSecret(path); err != nil {
			cfg, err = credentialprovider.ReadDockercfgFile(path)
		}
	}
	return cfg, err
}
```

Which is what we *used* to call, `credentialprovider.ReadDockerConfigJSONFile(path)` checks for `config.json` at https://github.com/openshift/builder/blob/master/vendor/k8s.io/kubernetes/pkg/credentialprovider/config.go#L121-L144

`ReadDockerConfigJsonFileGeneratedFromSecret(path)` calls `credentialprovider.ReadSpecificDockerConfigJsonFile(filepath.Join(filePath, DockerConfigJsonKey))` where that is `DockerConfigJsonKey` is `.dockerconfigjson`

and the final `credentialprovider.ReadDockercfgFile(path)` at https://github.com/openshift/builder/blob/master/vendor/k8s.io/kubernetes/pkg/credentialprovider/config.go#L90-L117 looks for `.dockercfg`

I'm following that flow, but rather than getting a `DockerConfig`, I'm getting the fully qualified path of the file found.  We ultimately set that in `AuthFilePath`

gabemontero:
looks like pull authentication is ok, but push authentication isn't, when looking at e2e-aws

will circle back to some more local testing tomorrow

gabemontero:
OK I couldn't help it, I looked tonight.

Yeah, the credentialproviders.GetAuthentication is not able to find the Auth from our `AuthFilePath` ... which ultimately flows through `findAuthentication` ... the lookup logic is different than what was previously done via the `BasicDockerKeyring.Lookup` to get the single `AuthConfiguration` based on the image's registry server.

This is occuring with the internal image registry references.

Now, @mtrmac did mention in prior comments (https://github.com/openshift/builder/pull/102#discussion_r328924879) that pushing doesn't go to mirrors anyway and a single auth entry would suffice.

Gonna apply that approach.

gabemontero:
Now worried and push/pull secrets in general and the containers/image code having trouble finding the image-registry references.

We'll see how the builds where the pullImage is an imagestreamtag as well do.

We may have to fall back on building AuthConfigurations from the push/pull secrets and employing the normalizations done in containers/image

gabemontero:
yeah use of the pull secret and imagestreamtags isn't working based on the e2e-aws run

the other runs just missed on the https://github.com/openshift/origin/pull/23891 merging
let's re-run those to filter out the `oc exec -p` noise

/test e2e-aws-image-ecosystem
/test e2e-aws-builds

gabemontero:
OK I got the `SystemContext.AuthFilePath` form to work with a change to the containers/image vendored in.

The legacy path check in `GetAuthentication` does not account for the fact that the `.dockercfg`  file may reside somewhere other than the home dir.

Looking at the latest containers/image code, the situation still exists:  https://github.com/containers/image/blob/master/pkg/docker/config/config.go#L87-L101

@mtrmac - my preference is to update containers/image so we can continue to use `SystemContext.AuthFilePath` and not have to employ normalization logic on the openshift/builder side

if you concur, I'll open a containers/image PR and reach out to you the assist with reviewing/merge (I'm assuming cutting a new github tag is a more drawn out process).

gabemontero:
ok image eco system is now passing with the containers/image patch @mtrmac @adambkaplan when it was previously failing 

good sign

gabemontero:
Opened https://github.com/containers/image/pull/722 @mtrmac @adambkaplan

gabemontero:
ok all green - good 

pushing version 2 base of review of https://github.com/containers/image/pull/722 with @mtrmac

gabemontero:
round 2 has passed .... porting to the c/image PR

pkg/build/builder/cmd/dockercfg/cfg.go
mtrmac: You’re completely right, I don’t know how I missed the two other cases.

gabemontero: no worries - good we are on the same page

gabemontero:
ok @adambkaplan @nalind @mtrmac @openshift/openshift-team-developer-experience FYI / PTAL

We now have
a) the same openshift/builder change that @mtrmac and I had settled 
b) fyi @adambkaplan 's earlier comments I believe have been rendered inapplicable as a result of subsequent change from a)
c) pulling the containers/image fix I submitted so that it can understand/handle legacy docker config formats like the one our build image pull secret uses
d) the switch from glide to gomod needed to make c) happen, as containers/image introduced go mod versioning as part of going to its v4
e) a bump to containers/buildah was needed as well as a result of d)

Let's see if we get the same success pulling from the image registry with the "official" containers/image fix that we got with my patched version last week.

mtrmac:
ACK — although I haven’t reviewed the migration to go.mod itself. The -5M lines look great but scary.

gabemontero:
Forgot to mention I left the glide.* files in for now, just as a reference, until the PR/regression testing fleshes out.

gabemontero:
Yeah the massive vendor churn would appear to be making things unhappy .... a few of these minimally:

`"openshift-git-clone flag redefined: log_dir\npanic: openshift-git-clone
            flag redefined: log_dir\n\ngoroutine 1 [running]:\nflag.(*FlagSet).Var(0xc42009a060,
            0x1a73760, 0x271cad0, 0x1891cbc, 0x7, 0x18d973b, 0x2f)\n\t/usr/local/go/src/flag/flag.go:810
            +0x540\nflag.(*FlagSet).StringVar(0xc42009a060, 0x271cad0, 0x1891cbc,
            0x7, 0x0, 0x0, 0x18d973b, 0x2f)\n\t/usr/local/go/src/flag/flag.go:713
            +0x8b\ngithub.com/openshift/builder/vendor/k8s.io/klog.InitFlags(0xc42009a060)\n\t/go/src/github.com/openshift/builder/vendor/k8s.io/klog/klog.go:420
            +0x8b\ngithub.com/openshift/builder/vendor/k8s.io/component-base/logs.init.0()\n\t/go/src/github.com/openshift/builder/vendor/k8s.io/component-base/logs/logs.go:36
            +0x2d\n"`

gabemontero:
ok starting over I got a gomod based builder image with c/images at v3.0.2 and c/buildah at v1.10.3 to work in my local env.

Gonna try more incremental changes from there, see where things land.

gabemontero:
OK these series of commits works for me locally ... none of the weird klog errors 

let's see how the CI jobs do @adambkaplan @mtrmac

gabemontero:
OK e2e-aws and e2e-aws-image-ecosystem passed ... good sign

gabemontero:
Hmmmm, aside from OOMKilled flake, maybe a single IST based image pulled that failed with an auth error ?

`[Feature:Builds][Slow] build can have container image source buildconfig with input source image and s2i strategy should complete successfully and contain the expected file [Suite:openshift]` from test/extended/builds/image_source.go

Why only a single one?  Will look at it / run it locally tomorrow AM.

Otherwise e2e-aws-builds clean

gabemontero:
OK runs for me locally ... perhaps a flake ... will retry before pushing debug changes

/retest

gabemontero:
yep was a flake ... all e2e's passing

@adambkaplan PTAL .... this is can now be considered for merge

quick note on the commits - I'd like to keep them broken up like they currently are if you agree.  I think showing the gomod evolution as I progressed through the various stages has instructive value.

go.mod
```diff
@@ -0,0 +1,81 @@
+module github.com/openshift/builder
+
+go 1.12
+
+require (
+	github.com/MakeNowJust/heredoc v0.0.0-20170808103936-bb23615498cd
+	github.com/certifi/gocertifi v0.0.0-20180905225744-ee1a9a0726d2 // indirect
+	github.com/containerd/continuity v0.0.0-20190426062206-aaeac12a7ffc // indirect
+	github.com/containers/buildah v1.11.3
+	github.com/containers/image v3.0.2+incompatible
```
mtrmac: I would expect this line to go away or be an `// indirect` dependency.

After making the two other edits, `go mod vendor` removes the non-v4 copy.

pkg/build/builder/daemonless.go
```diff
@@ -13,9 +13,9 @@ import (
 	"github.com/containers/buildah/imagebuildah"
 	"github.com/containers/buildah/util"
 	ireference "github.com/containers/image/docker/reference"
```
mtrmac: This was not moved to `/v4/`. (Might work fine, not sure.)

pkg/build/builder/dockerutil.go
```diff
@@ -32,7 +32,7 @@ type DockerClient interface {
 	PushImage(opts docker.PushImageOptions, auth docker.AuthConfiguration) (string, error)
```
mtrmac: Above, import of `idocker "github.com/containers/image/docker"` was not moved to `/v4/`. This breaks `unwrapUnauthorizedError` if it receives an error from `/v4/` implementations.

mtrmac:
> quick note on the commits - I'd like to keep them broken up like they currently are if you agree. I think showing the gomod evolution as I progressed through the various stages has instructive value.

Do the commits move from one working state to another? I expected at least the one updating Buildah to use /v4/ without updating any of the callers in this project to break things.

pkg/build/builder/cmd/dockercfg/cfg.go
```diff
@@ -103,6 +128,48 @@ func GetDockercfgFile(path string) string {
...
+			if err != nil {
+				log.V(4).Infof("while trying to canonicalize %s: %v", configPath, err)
+				continue
+			}
+			klog.V(4).Infof("looking for %s at %s", file, absDockerConfigFileLocation)
```
adambkaplan: nit - why is this using `klog` instead of `log`?

gabemontero:
> BTW, `go mod tidy` fails with
> 
> ```
> go: import "github.com/openshift/builder/pkg/build/builder" ->
> 	import "github.com/openshift/source-to-image/pkg/build/strategies" ->
> 	import "github.com/openshift/source-to-image/pkg/build/strategies/onbuild" ->
> 	test ->
> 	import "github.com/docker/docker/builder/dockerfile/parser": looping trying to add package
> ```
> 
> Not sure whether that’s known or worth worrying about.

I explicitly reframed from `go mod tidy` this time, as my prior incarnation, where I used it, resulted in the `klog` related panic in k8s on startup of the build containers in openshift.

After merge, and when `v4.1.0` of c/image is available, we can revisit whether a pass with `go mod tidy` does not subvert things.

gabemontero:
> > quick note on the commits - I'd like to keep them broken up like they currently are if you agree. I think showing the gomod evolution as I progressed through the various stages has instructive value.
> 
> Do the commits move from one working state to another? 
yes
I expected at least the one updating Buildah to use /v4/ without updating any of the callers in this project to break things.

Yep that is true.  There is a subsequent commit that updates pkg/build/builder/* where needed to change the imports to c/image to c/image/v4.

Pros vs. cons, IMO, that is the lesser concern.

mtrmac:
I meant whether every commit as a single unit “from one working state to another”.

OTOH it does make sense to separate the large vendoring commits from the small code change commits that should actually be reviewed.

It’s up to this project to decide which one is more important, just noting that there is this tension.

pkg/build/builder/daemonless.go
gabemontero: thanks for catching this .... updated locally /testing now

pkg/build/builder/dockerutil.go
gabemontero: thanks for catching this ... updated locally / testing now

mtrmac:
… moving from one working state to another can be useful for things like `git bisect`. OTOH I’ve seen that much more frequently talked about than used :)

pkg/build/builder/daemonless.go
gabemontero: update works locally ... have pushed

pkg/build/builder/dockerutil.go
gabemontero: update works locally ... have pushed

pkg/build/builder/cmd/dockercfg/cfg.go
gabemontero: will fix momentarily

gabemontero: update pushed @adambkaplan thanks

go.mod
mtrmac: Looking at e7c2cd6b6ff6f7d261c27c92a967e4cb19ec85ea , please run `go mod vendor` as well.

gabemontero: I ran `go mod vendor` multiple times through the process.

Do you mean run `go mod vendor` on top of my last commit here?

mtrmac: Yes, on top of the one that removes the last `/v4/` bits. (Or, being pedantic, that part should be moved from the “seed containers/image with…” commit into the  move containers/image imports to v4” one.

After checking out the commit I quoted above, `go mod vendor` removes `vendor/github.com/{containers/image,docker/libtrust}` for me.

gabemontero: thanks for clarifying ... did for me as well

and things work locally ... will push momentarily

gabemontero:
OOM Kill flake last e2e-aws-build run

gonna push a non-vendor commit squash, initiate re-run

gabemontero:
Interesting ... the `git log -5` on my laptop shows a different order than the commit view in the PR (though the commit hash IDs are the same:

```
gmontero ~/go/src/github.com/openshift/builder  (disconnected-more-auth)$ git log -5
commit a1a5b06acb8df8fb06fd5f44703752ecd6450ae8 (HEAD -> disconnected-more-auth, origin/disconnected-more-auth)
Author: gabemontero <gmontero@redhat.com>
Date:   Mon Oct 7 20:33:00 2019 -0400

    seed containers/image with entire dockerconfig for authentication in case mirrors are in play; move c/image imports to v4

commit b626e817b12e21f6c073c0e6c6a17ad01a477394
Author: gabemontero <gmontero@redhat.com>
Date:   Tue Oct 8 14:24:41 2019 -0400

    one more go mod vendor to remove pre-v4 c/image files

commit b8760a0775903e40b7120ebae6087dcead4c889d
Author: gabemontero <gmontero@redhat.com>
Date:   Mon Oct 7 20:46:49 2019 -0400

    bump to my containers/image commit

commit ee844b9a87fc0b99a2cb65c28dc1feb4181dfdeb
Author: gabemontero <gmontero@redhat.com>
Date:   Mon Oct 7 20:24:57 2019 -0400

    bump container/image, container/buildah

commit d4b1c3fd5b3dfd5c30b9b230c791c2904526dbb6
Author: gabemontero <gmontero@redhat.com>
Date:   Mon Oct 7 20:15:33 2019 -0400

    switch vendor to gomod
```

adambkaplan:
@gabemontero I believe GitHub show commits by timestamp, not hash ordering

adambkaplan:
@gabemontero the OOM flake may be a red herring. It's known to flake, but it often passes the second time around.

Early after the rebase I found a flake on the ocm's daemonset rollout timing out. This IMO is a real performance regression. I believe there is a BZ for it.

gabemontero:
thanks for the comebacks @adambkaplan 

we are green again with the e2e's .... can we merge with the gomod/vendor commits in their current form?

adambkaplan:
@gabemontero for the sake of documenting a "real world" go mod migration, I recommend taking what you have as a new branch and pushing it to your fork. That branch + a gist or Google doc walking through what you did would be very beneficial.

For actual code history I prefer we squash the net results.

gabemontero:
squashed down to a vendor commit and code commit @adambkaplan per the usual pattern

PTAL / tag as you deem appropriate - thanks

gabemontero:
aws/terraform flake with e2e-aws

gabemontero:
e2e-aws-builds has the oom ds rollout flake, a build did not get scheduled cause the cluster is constrained (where I see a 3 dozen MCO experiencing throttling events at the same time in the cluster events.json), and a problem committing a built image on the third:

`2019-10-10T14:48:08.546999633Z error: error creating buildah builder: Error committing the finished image: error adding layer with blob "sha256:caa1771d2710f104d1441da00e8b236c3ba9cde637dc60c066bdb99ebe40412b": Error processing tar file(io: read/write on closed pipe)`

@nalind @dmage : does the above message ring any alarms for you guys?

In total, none of those are related to this PR.

/test e2e-aws-builds

gabemontero:
ok e2e-aws got past aws/terraform flakes

mtrmac:
> and a problem committing a built image on the third:
> 
> `2019-10-10T14:48:08.546999633Z error: error creating buildah builder: Error committing the finished image: error adding layer with blob "sha256:caa1771d2710f104d1441da00e8b236c3ba9cde637dc60c066bdb99ebe40412b": Error processing tar file(io: read/write on closed pipe)`

FWIW this did show up in Slack on Aug 9th (with `podman pull`), but the cause wasn’t found that time (the user deleted the storage and a second attempt worked).

gabemontero:
2 of the 3 non-related flakes (minus the buildah one) as the last e2e-aws-builds 

/retest

gabemontero:
we are all green again @adambkaplan for this build/disconnected PR

pkg/build/builder/daemonless.go
```diff
@@ -267,9 +248,6 @@ func pushDaemonlessImage(sc types.SystemContext, store storage.Store, imageName
 	}
 
 	systemContext := sc
-	// if credsDir, ok := os.LookupEnv("PUSH_DOCKERCFG_PATH"); ok {
-	// 	systemContext.AuthFilePath = filepath.Join(credsDir, ".docker", "config.json")
-	// }
 	systemContext.AuthFilePath = "/tmp/config.json"
```
adambkaplan: @gabemontero did we miss adding the same logic for image push? Granted, pushing to a mirror feels like a bad use case to support.

pkg/build/builder/daemonless.go
```diff
@@ -536,12 +514,12 @@ func (d *DaemonlessClient) RemoveContainer(opts docker.RemoveContainerOptions) e
 	return err
 }
 
-func (d *DaemonlessClient) PullImage(opts docker.PullImageOptions, auth docker.AuthConfiguration) error {
+func (d *DaemonlessClient) PullImage(opts docker.PullImageOptions, searchPaths []string) error {
```
adambkaplan: ditto - do we need to do the same for image push?

pkg/build/builder/daemonless.go
gabemontero: It got discussed @adambkaplan and was deemed unnecessary.

pkg/build/builder/daemonless.go
gabemontero: It got discussed @adambkaplan and was deemed unnecessary.

pkg/build/builder/daemonless.go
gabemontero: In the end I decided to minimize the moving parts

gabemontero: for reference:  https://github.com/openshift/builder/pull/102#discussion_r328924879

pkg/build/builder/daemonless.go
gabemontero: for reference:  https://github.com/openshift/builder/pull/102#discussion_r328924879

gabemontero:
e2e-aws-image-ecosystem had a non-related security flake that prevented a pod from getting scheduled:

`Message: "pods \"test-pod-e1be847d-9d98-4625-a247-eec2cebd9fd1\" is forbidden: no SecurityContextConstraints found in cluster",`

/test e2e-aws-image-ecosystem

gabemontero:
ok this time e2e-aws-image-ecosystem tests passed:  `2019/10/14 16:28:47 Container test in pod e2e-aws-image-ecosystem completed successfully` 

and all the image_ecosystem testcases are green, but some CI post processing failed to get a lease or something

/test e2e-aws-image-ecosystem

PR #106, opened 2019-10-29
Bug 1758305: Bump s2i

gabemontero:

/assign @adambkaplan


go.mod
```diff
@@ -74,6 +74,7 @@ replace (
 	github.com/containers/image/v4 => github.com/containers/image/v4 v4.0.0-20191004175620-9da78d1f50d2
 	github.com/docker/docker => github.com/openshift/moby-moby v1.4.2-0.20190308215630-da810a85109d
 	github.com/opencontainers/runtime-tools => github.com/opencontainers/runtime-tools v0.8.0
+	github.com/openshift/source-to-image => github.com/openshift/source-to-image v1.2.1-0.20191029175504-f20322e54448
```
adambkaplan: @gabemontero was this the result of `go get github.com/openshift/source-to-image@master`?

gabemontero: not exactly @adambkaplan , what I did was add a `replace` item in `go.mod` to force source-to-image to my commit that just merged 

the `go mod vendor` then gets "creative" IMO in how it updates both go.mod and go.sum to access that commit 

the result is what we wanted ... just the delta between the release you recently cut and my change

adambkaplan: You get the result, but imo it's an abuse of the `replaces` system.

If you revert, then do `GO111MODULE=on go get github.com/openshift/source-to-image@master`, go mod will update the version of s2i in the `requires` part of `go.mod`. A follow up `go mod tidy && go mod vendor` should pull everything in properly.

PR #108, opened 2019-10-30
move to containers/image v5, buildah v1.11.4

nalind:
Move to github.com/containers/image v5 and buildah v1.11.4. Builds which specify a base image using a digest or tag which points to a manifest list should not produce errors when pulling the base image, otherwise there should be no visible changes to behavior.

Includes a proposed update to update-deps.sh which more or less does what make vendor does in the buildah source tree.


mrunalp:
@adambkaplan ptal

mrunalp:
@bparees ptal

adambkaplan:
@nalind is there a BZ associated with the "build fails if FROM uses a mainfiest list"? I assume there is a buildah-related issue out there tracking this fix.

nalind:
I'm not sure we were tracking a problem in builds - one of the main motivations was https://github.com/cri-o/cri-o/issues/2157, along with other issues elsewhere tracking the `Digest of source image's manifest would not match destination reference` error at run-time.  It actually looks like we still have a bit of work when using a base that we specify using a digest.

nalind:
Additional work in https://github.com/containers/buildah/pull/1960 should be relevant.

PR #109, opened 2019-11-01
Bump docker/swarmkit and remove logrus

dmage:


adambkaplan:
@dmage is swamkit the only lib that used the wrong name for logrus?

dmage:
@adambkaplan no vendored dependencies use the wrong name. But swarmkit was the last dependency that tried to bring the wrong name to go.mod.

PR #116, opened 2019-11-20
Bug 1774492: turn off forcepull after the special incremental/forcepull pre-build …

gabemontero:
…one off

/assign @adambkaplan 
/assign @bparees 

this replaces the OCM PR I just closed.

### DIFF
--- a/pkg/build/builder/sti.go
+++ b/pkg/build/builder/sti.go
@@ -251,6 +251,11 @@ func (s *S2IBuilder) Build() error {
 				log.V(5).Infof("Incremental image pull failure: %v", err)
 				config.Incremental = false
 				config.IncrementalFromTag = ""
+			} else {
+				// with the 'special case pull' of the incremental image above, set force pull to false so that the
+				// logic below does not attempt to re-pull the previous-output-image with the standard pull secret
+				// (which would potentially fail since the image should be pulled w/ the push credentails)
+				s.build.Spec.Strategy.SourceStrategy.ForcePull = false
 			}
 		}
 	}


gabemontero:
when reviewing the bug I would suggest expanding the view so you see up to line 240 of the changed file @adambkaplan @bparees

gabemontero:
verified locally via

1) used BC of 

apiVersion: build.openshift.io/v1
kind: BuildConfig
metadata:
  name: sti
spec:
  output:
    to:
      kind: ImageStreamTag
      name: sti:latest
  postCommit: {}
  resources: {}
  runPolicy: Serial
  source:
    binary: {}
    type: Binary
  strategy:
    sourceStrategy:
      forcePull: true
      from:
        kind: DockerImage
        name: registry.redhat.io/openjdk/openjdk-11-rhel8:latest
      incremental: true
    type: Source

2) created a secret for my registry.redhat.io credentials
3) linked the secret to the default and builder SAs
4) ran the build from my local clone of the openshift jenkins plugin maven/java based repo (as the binary build input)

and extended test following a similar pattern and using the installer pull creds seems possible, where we clone a java repo and then do a binary build from it ... or we come up with a non-binary build example.

bparees:
/hold

i want to better understand this first

pkg/build/builder/sti.go
```diff
@@ -251,6 +251,10 @@ func (s *S2IBuilder) Build() error {
 				log.V(5).Infof("Incremental image pull failure: %v", err)
 				config.Incremental = false
 				config.IncrementalFromTag = ""
+			} else {
+				// turn off force pull so that we do not attempt to re-pull with the pull creds in the builder.Build
+				// call below
+				s.build.Spec.Strategy.SourceStrategy.ForcePull = false
```
bparees: among other things, wouldn't this still break for a "normal" s2i build w/ forcePull set?

gabemontero: When I look at line 240 here @bparees and the `if config.Incremental` check it makes me thing that this would not apply to a "normal" s2i build with forcePull set.

Or am I misunderstanding what you mean by "normal" s2i build with forcePull set.

bparees: my concern was that the failure would still occur for a non-incremental build.  but i think i now understand that the failure is caused by buildah attempting to pull the prev-output image (since it's listed in the Dockerfile), using the normal pull creds, and thus failing, so the failure mode would not be applicable to a non-incremental build.

gabemontero: comment updated per conversation with @bparees on slack

gabemontero:
terraform / aws hoopla for e2e-aws and e2e-aws-image-eco

gabemontero:
same with the e2e-aws-builds

PR #113, opened 2019-11-13
bump github.com/containers/buildah to v1.11.5

nalind:
Bump buildah from v1.11.4 to v1.11.5.

Also, use -mod=vendor when building in hack/build.sh, so that it won't try to update dependencies, since that's hack/update-deps.sh's job.

PR #119, opened 2019-12-09
Bug 1772179: bump github.com/containers/{storage,buildah}

nalind:
Bump github.com/containers/storage to 4822c60538402940cb8044c96659a46ffbab3bd1, between v1.15.2 and v1.15.3, to pick up a fix for a race in the library.

Bump github.com/containers/buildah to 8d6869bccee02f73f0e26e739c9f8712578b3f32, modifying how we unmount the container's rootfs when using chroot isolation so that we just detach the whole thing rather than trying to detach each submount in order, which isn't necessary.

This is https://bugzilla.redhat.com/show_bug.cgi?id=1772179, but for 4.4/master.


nalind:
This will obsolete #113.

go.mod
```diff
@@ -52,18 +51,20 @@ require (
...
-	golang.org/x/sys v0.0.0-20190924154521-2837fb4f24fe
+	golang.org/x/sys v0.0.0-20191127021746-63cb32ae39b2
 	k8s.io/api v0.0.0
 	k8s.io/apimachinery v0.0.0
 	k8s.io/client-go v0.0.0
 	k8s.io/component-base v0.0.0
 	k8s.io/klog v1.0.0
 	k8s.io/kubectl v0.0.0
-	k8s.io/kubernetes v0.0.0
+	k8s.io/kubernetes v1.13.0
```
adambkaplan: @nalind please revert this - we have a `replace` stanza where we set this to `v1.17.0` (using replace right now because the release candidate did not have a valid semver).

For what it's worth it appears that no kube bumps were changed.

nalind: These are the values `update-deps.sh` changes them to.  Are you sure?

adambkaplan: if this is coming from `update-deps.sh`, let's keep it :/

adambkaplan:
/retitle Bug 1772179: bump github.com/containers/{storage,buildah}

go.mod
nalind: Okay, then.

PR #122, opened 2019-12-10
Bug 1782409: bump github.com/containers/{storage,buildah} 

nalind:
Bump github.com/containers/storage to 4822c60538402940cb8044c96659a46ffbab3bd1, between v1.15.2 and v1.15.3, to pick up a fix for a race in the library.

Bump github.com/containers/buildah to 8d6869bccee02f73f0e26e739c9f8712578b3f32, modifying how we unmount the container's rootfs when using chroot isolation so that we just detach the whole thing rather than trying to detach each submount in order, which isn't necessary.

This is #119, but for 4.3 specifically.


adambkaplan:
/retitle Bug 1782409: bump github.com/containers/{storage,buildah}

TomSweeneyRedHat:
Addresses: https://bugzilla.redhat.com/show_bug.cgi?id=1782409

PR #123, opened 2019-12-11
bump github.com/containers/storage and dependencies for go 1.13

nalind:
Update to using the results of using go 1.13 for dependency management, and building using `-mod=vendor`.  This rolls us back to an older version of `libnetwork`, but since we don't use it to set up any networking in the build, I don't expect any side-effects.

`go mod` and `update-deps.sh` left the `go 1.12` mention in `go.mod` unchanged, but it doesn't complain if we manually change it to `go 1.13`.  I can adjust that if it's desired.

Also updates github.com/containers/storage to v1.15.3.


nalind:
We're still building with 1.12, and I don't see 1.13 versions of our base images, so `build.sh` needs to make its use of `-mod=vendor` conditional.

rhatdan:
LGTM

go.mod
```diff
@@ -62,10 +62,14 @@ require (
 )
 
 replace (
-	github.com/containers/storage => github.com/containers/storage v1.13.3-0.20191209173733-4822c6053840
+	github.com/containerd/containerd => github.com/containerd/containerd v0.2.10-0.20170808145631-06b9cb351610
```
adambkaplan: @nalind I had issues when I using newer versions of containerd. Have you found the same?

hack/lib/build/binaries.sh
```diff
@@ -3,11 +3,25 @@
...
 # os::build::build_binary builds the openshift-builder binary
 function os::build::build_binary() {
   # Fetch the version.
   local version_ldflags
   version_ldflags=$(os::build::ldflags)
-  go build -ldflags "${version_ldflags}" -tags "${OS_GOFLAGS_TAGS-}" -o openshift-builder ./cmd
+  # Fetch additional build flags.
+  local mod_vendor_flag
+  mod_vendor_flag=$(os::build::modvendorflag)
+  go build ${mod_vendor_flag} -ldflags "${version_ldflags}" -tags "${OS_GOFLAGS_TAGS-}" -o openshift-builder ./cmd
```
adambkaplan: Note that on CI the golang 1.13 imagestream has the `-mod=vendor` flag set.

go.mod
nalind: I did, yes.

hack/lib/build/binaries.sh
nalind: Not sure I understand this.  Do you have a pointer to how it's done?  Is it done by setting GOFLAGS?  Should I drop this chunk?

adambkaplan: > Is it done by setting GOFLAGS?

Yes - I don't believe that flag is set in our golang 1.12 imagestream. In 1.13 `GOFLAGS="-mod=vendor` is set as an env var because of a golang bug. I don't see any harm setting it here in though.

go.mod
adambkaplan: @nalind how did you choose this specific version? It seems this refers to a SHA that is not associated with a tag, nor is it the latest commit in the `v0.2.x` branch.

hack/lib/build/binaries.sh
nalind: Makes sense.  The option isn't recognized by 1.12.

go.mod
nalind: It was the version referenced by the `vendor.conf` in the version of `github.com/docker/docker` tree that we're using.  But you're right that there are later commits in the `v0.2.x` branch, and moving to its tip (`408d13de2fbb2298a9ae7c14d86803a740ab12a9`) doesn't appear to break compilation.  Should I update to point to that instead?

adambkaplan: @nalind let's do that, and add a comment that we're using the tip of that branch (ex: `# Result of go get github.com/containerd/containerd@v0.2.x`)

nalind: Okay, going with that.

adambkaplan:
/hold

Build tests are broken due to https://bugzilla.redhat.com/show_bug.cgi?id=1785326.

PR #130, opened 2020-02-10
Bug 1801388: use of ARG step in Dockerfile causes panic in imagebuilder (openshift…

gabemontero:
…/builder suppplies bad parameter)

/assign @adambkaplan 
@Rulox FYI

### DIFF
--- a/pkg/build/builder/common.go
+++ b/pkg/build/builder/common.go
@@ -448,7 +448,7 @@ func replaceImagesFromSource(node *parser.Node, imageSources []buildapiv1.ImageS
 		}
 	}
 	names := make(map[string]string)
-	stages, err := imagebuilder.NewStages(node, imagebuilder.NewBuilder(nil))
+	stages, err := imagebuilder.NewStages(node, imagebuilder.NewBuilder(make(map[string]string)))
 	if err != nil {
 		return err
 	}
@@ -488,7 +488,7 @@ func findReferencedImages(dockerfilePath string) ([]string, error) {
 	}
 	names := make(map[string]string)
 	images := sets.NewString()
-	stages, err := imagebuilder.NewStages(node, imagebuilder.NewBuilder(nil))
+	stages, err := imagebuilder.NewStages(node, imagebuilder.NewBuilder(make(map[string]string)))
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/build/builder/common_test.go
+++ b/pkg/build/builder/common_test.go
@@ -199,6 +199,19 @@ func Test_addBuildParameters(t *testing.T) {
 		},
 		{
 			original: heredoc.Doc(`
+				ARG GOLANG_CONTAINER=golang:latest
+				FROM $GOLANG_CONTAINER
+				RUN echo "hello world"
+				`),
+			want: want{
+				Out: heredoc.Doc(`
+				FROM $GOLANG_CONTAINER
+				RUN echo "hello world"
+				`),
+			},
+		},
+		{
+			original: heredoc.Doc(`
 				FROM scratch
 				# FROM busybox
 				RUN echo "hello world"


pkg/build/builder/common_test.go
```diff
@@ -197,6 +197,19 @@ func Test_addBuildParameters(t *testing.T) {
...
 		},
+		{
+			original: heredoc.Doc(`
+				ARG GOLANG_CONTAINER=golang:latest
+				FROM $GOLANG_CONTAINER
+				RUN echo "hello world"
+				`),
+			want: want{
+				Out: heredoc.Doc(`
+				FROM $GOLANG_CONTAINER
```
gabemontero: Note the actual substitution does not happen as part of the method we call below.

See https://github.com/openshift/builder/blob/master/vendor/github.com/openshift/imagebuilder/dispatchers.go#L519-L523

gabemontero:
install failure image-eco

gabemontero:
/hold

the unit test not hitting the panic does not appear to be sufficient coverage

the centos mirrors are up again, was able to build the builder image, and am still seeing what looks like a panic (maybe in a different spot)

gabemontero:
> /hold
> 
> the unit test not hitting the panic does not appear to be sufficient coverage
> 
> the centos mirrors are up again, was able to build the builder image, and am still seeing what looks like a panic (maybe in a different spot)

Check that - I did not properly update the builder image my cluster is using.

Now doing so, the panic is gone, but the substitution is not happening.

`error: build error: error parsing name "$GOLANG_CONTAINER": error parsing image name "$GOLANG_CONTAINER": invalid reference format: repository name must be lowercase
`

This could be an imagebuilder issue ... investigating.

pkg/build/builder/common_test.go
gabemontero: Actually something may be amiss here.  See below.

gabemontero:
> > /hold
> > the unit test not hitting the panic does not appear to be sufficient coverage
> > the centos mirrors are up again, was able to build the builder image, and am still seeing what looks like a panic (maybe in a different spot)
> 
> Check that - I did not properly update the builder image my cluster is using.
> 
> Now doing so, the panic is gone, but the substitution is not happening.
> 
> `error: build error: error parsing name "$GOLANG_CONTAINER": error parsing image name "$GOLANG_CONTAINER": invalid reference format: repository name must be lowercase `
> 
> This could be an imagebuilder issue ... investigating.

Actually to support ARG substitution for multiple FROMs looks like will require changes in openshift/builder, adding more imagebuilder calls.

I'll see if I can sort out a solution @adambkaplan but depending on how much of a rabbit hole it becomes we might want to classify any post panic fixes as features, and add doc/readme updates around this.

Will report back after giving it a go some more.

gabemontero:
> > > /hold
> > > the unit test not hitting the panic does not appear to be sufficient coverage
> > > the centos mirrors are up again, was able to build the builder image, and am still seeing what looks like a panic (maybe in a different spot)
> > 
> > 
> > Check that - I did not properly update the builder image my cluster is using.
> > Now doing so, the panic is gone, but the substitution is not happening.
> > `error: build error: error parsing name "$GOLANG_CONTAINER": error parsing image name "$GOLANG_CONTAINER": invalid reference format: repository name must be lowercase `
> > This could be an imagebuilder issue ... investigating.
> 
> Actually to support ARG substitution for multiple FROMs looks like will require changes in openshift/builder, adding more imagebuilder calls.
> 
> I'll see if I can sort out a solution @adambkaplan but depending on how much of a rabbit hole it becomes we might want to classify any post panic fixes as features, and add doc/readme updates around this.
> 
> Will report back after giving it a go some more.

OK I have *something* that works @adambkaplan 

Will look at it with fresh eyes tomorrow and see what spit/polish is needed to make it presentable.

gabemontero:
OK after an exchange with @bparees [on slack](https://coreos.slack.com/archives/CE2HALN2W/p1581446719378100) we should not have to deal with `FROM $GOLANG_CONTAINER AS BUILDER` in https://github.com/nginxinc/kubernetes-ingress/blob/master/build/Dockerfile in openshift/builder ... `buildah` should pick up the default and we should not fail with  https://github.com/openshift/builder/pull/130#issuecomment-584442318

So this fix for the panic should suffice, and I'll clone bug 1801388 and send it to buildah to let us know if this is a new or existing bug, etc. etc.

/retest

pkg/build/builder/common.go
```diff
@@ -448,7 +448,7 @@ func replaceImagesFromSource(node *parser.Node, imageSources []buildapiv1.ImageS
 		}
 	}
 	names := make(map[string]string)
-	stages, err := imagebuilder.NewStages(node, imagebuilder.NewBuilder(nil))
+	stages, err := imagebuilder.NewStages(node, imagebuilder.NewBuilder(names))
```
nalind: The current version of imagebuilder uses this map by reference, which will conflict with our use of it below.  Can we use a fresh map here, and on line 491?

gabemontero: So for line 491 I see `stages, err := imagebuilder.NewStages(node, imagebuilder.NewBuilder(nil))` so I am interpreting the need to pass a non-nil map into that call as well.... granted that is in a different method, `findReferencedImages`

@nalind  - If you are asking for changes around the map within `replaceImagesFromSource` please clarify

gabemontero: OK I see what you did in https://github.com/openshift/builder/pull/131/files ... I presume you mean that.

gabemontero: updated @nalind thanks (I see the use map by reference element now :-) ) !!

pkg/build/builder/common.go
```diff
@@ -448,7 +448,7 @@ func replaceImagesFromSource(node *parser.Node, imageSources []buildapiv1.ImageS
 		}
 	}
 	names := make(map[string]string)
-	stages, err := imagebuilder.NewStages(node, imagebuilder.NewBuilder(nil))
+	stages, err := imagebuilder.NewStages(node, imagebuilder.NewBuilder(make(map[string]string)))
```
bparees: the prior code only threw an error on
```
ARG foo=bar
FROM centos
```

it did not throw an error on:
```
FROM centos
ARG foo=bar
```

so i'd like to better understand why this is the right fix before we merge it.

bparees: also, we normally call both `findReferencesImages` and `replaceImagesFromSource` (in that order).

it's the second call that fails.

but both functions invoke:
` imagebuilder.NewStages(node, imagebuilder.NewBuilder(make(map[string]string)))`

so why does the second invocation fail and not the first?  Some stateful mutation of `node`?

gabemontero: So the panic occurred at https://github.com/openshift/builder/blob/master/vendor/github.com/openshift/imagebuilder/dispatchers.go#L557

With `arg` method, the `Builder` object ptr passed in corresponds to the `imagebuilder.NewBuilder` method call here, where the map gets set to the `Args` field in the `Builder` object.

The call to `NewStages` leads to a call on this function (which is a function on the `Builder` created by `imagebuild.NewBuilder`):

```
func (b *Builder) extractHeadingArgsFromNode(node *parser.Node) error {
	var args []*parser.Node
	var children []*parser.Node
	extract := true
	for _, child := range node.Children {
		if extract && child.Value == command.Arg {
			args = append(args, child)
		} else {
			if child.Value == command.From {
				extract = false
			}
			children = append(children, child)
		}
	}

	for _, c := range args {
		step := b.Step()
		if err := step.Resolve(c); err != nil {
			return err
		}
		if err := b.Run(step, NoopExecutor, false); err != nil {
			return err
		}
	}

	node.Children = children
	return nil
}
```

The `step` in that final for loop corresponds to the `ARG` step in the Dockerfile.  And the `b.Run` call results in a call to `func arg(....)`.

I interpreted the need for having to supply the map for `imagebuilder.NewBuilder`, vs. that method creating the map, as analogous to `--build-args`with the various k/v pairs associated in the various ARGS steps.  If you had those, they would be in the map passed in.

At this, we are only filling in ARGS with have default values.  So the empty map suffices.

It did not throw the exception on the later case since the ARG is defined after the FROM, it cannot be used with the FROM.

I *think* that covers it.

nalind: https://github.com/openshift/imagebuilder/pull/147#issue-373860798 is where I'd like to fix the panic in imagebuilder, but I expect that this patch here would also prevent it.

gabemontero: it would .... and we could always do this short term, and then switch it after an imagebuilder merge and vendor bump here.

But longer term @nalind , we still need to reconcile in the slack thread between you, @bparees , and I the need to do the default value ARG substitution for FROMs in openshift/builder via imagebuilder vs. having imagebuildah.BuildDockerfile do it somehow.

bparees: > also, we normally call both findReferencesImages and replaceImagesFromSource (in that order).

i think i was wrong about this, sorry for the noise.

gabemontero: RE: the [2nd slack discussion thread](https://coreos.slack.com/archives/CE2HALN2W/p1581451537394100) @bparees gave @nalind a simplified reproducer which bypasses the panic fixed with this PRs changes and produces the issue with ARG defaults not showing up in FROM references:

```
apiVersion: build.openshift.io/v1
kind: BuildConfig
metadata:
  name: sample-build-docker-args-preset
spec:
  source:
    dockerfile: |-
      FROM centos
      ARG foo=centos
      FROM $foo
    type: Dockerfile
  strategy:
    dockerStrategy: {}
    type: Docker
```

which results in:
```
I0211 22:06:43.140914       1 builder.go:318] Starting Docker build from build config sample-build-docker-args-preset-17 ...
F0211 22:06:43.209605       1 helpers.go:114] error: build error: error parsing name "$foo": error parsing image name "$foo": invalid reference format
```

@nalind said he would start poking around at it.

I'll update the cloned BZ on the buildah queue with this.

gabemontero:
we have green tests

@bparees @nalind I'd rather merge this now vs. waiting for the imagebuilder merge bump

And sort out how the 
```
      ARG foo=centos
      FROM $foo
```

scenario gets sorted out in a separate PR.

Thoughts?

pkg/build/builder/common_test.go
```diff
@@ -197,6 +197,19 @@ func Test_addBuildParameters(t *testing.T) {
...
+				RUN echo "hello world"
+				`),
+			},
```
bparees: 1) why are we expecting this to strip the `ARG` directive from the dockerfile?

2) don't we need a test that proves we actually substitute the FROM line correctly? e.g.

```
FROM centos
ARG foo=bar
FROM $foo
```
in a buildconfig that has a dockerstrategy.From section?

gabemontero: These unit tests only deal with the imagebuilder translations prior to passing the dockerfile to buildah

If we expect buildah to handle the case above on its own, we would not add a test here.

Ultimately, when this is fixed, we'll need an builds e2e in openshift/origin like the ones you linked to yesterday that do a dockerfile like the one you've noted.

gabemontero: The unit test I added caused the panic before my fix.

It does not panic after my fix.

gabemontero:
> /lgtm
> 
> @gabemontero @nalind I've split out the subsequent ARG issue into a new BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1802202

@adambkaplan I've added a dependency to https://bugzilla.redhat.com/show_bug.cgi?id=1801875 to the bug you just opened, as @nalind currently has the to-do on understanding (and fixing if it makes sense) why the buildah api's are not using the default value for the ARG on its own, without requiring imagebuilder based translations in the openshift/builder layer.

We can use the bug you opened to either bump the buildah dependency with @nalind 's fix, or add imagebuilder based translations if he convinces us it can't be done in buildah.

pkg/build/builder/common_test.go
bparees: > These unit tests only deal with the imagebuilder translations prior to passing the dockerfile to buildah

ok, but why do we expect that translation to strip the ARG directive?


> If we expect buildah to handle the case above on its own, we would not add a test here.

not the same thing.  There are two ways for a FROM reference in a dockerfile to be replaced:

1) something in your buildconfig which tells us to replace specific FROM references with a resolved imagestream.  this has nothing to do w/ buildah and nothing to do with any ARGs

2) using an ARG reference in a FROM line in the dockerfile.  This is the substitution buildah should be doing (but is not handling correctly today in all cases)

I'm referring to case (1), not case (2).

That said, upon further reflection i don't know that we *need* to add (1) per se.  It might be nice to have (to ensure we can do a proper substitution even when the FROM value is "$foo" and not a valid pullspec) but it's not a bug/regression that is being fixed in this PR.

> Ultimately, when this is fixed, we'll need an builds e2e in openshift/origin like the ones you linked to yesterday that do a dockerfile like the one you've noted.

No, the e2e we'll need at that point would have to be one where we do *not* have a dockerstrategy From imagestream reference, so that the dockerfile FROM $foo gets substituted by buildah via the ARG reference, not by the openshift "replace the last FROM with the imagestream reference" logic.

nalind: I think the ARG is getting stripped in `ManageDockerfile()`.  The `replaceImagesFromSource()` function calls `imagebuilder.NewStages()`, which modifies the parsed tree(!), discarding any instructions that precede the first FROM instruction while doing so, so the reconstructed Dockerfile that it writes doesn't get to include them.

I'm working on a patch that pulls instructions from the parsed tree beforehand and then re-prepends them to the written Dockerfile to work around this.  It doesn't solve the substitutions-aren't-happening part of things, because we don't call into any imagebuilder functions that perform that resolution.

PR #132, opened 2020-02-12
Bug 1842982: preserve ARG values that come before the first FROM

nalind:
Some of the helper functions that `addBuildParameters()` calls modify the parsed tree, discarding `ARG` instructions that precede the first `FROM` instruction.  The replacement Dockerfile that we generate from that tree would be missing those instructions.

Knowing that this will happen, walk the parsed tree beforehand, recording `ARG` instructions that precede the first `FROM` instruction, and then re-prepend them to the Dockerfile that we generate from our tree after we've finished manipulating it.

If we've got `BuildArgs` in a `DockerStrategy`, populate the `BuildArgs` map that we pass to imagebuilder functions that might have a use for them.

We usually try to pull images that are referenced in the Dockerfile ourselves, so that we can record the time required to pull them.  If we looked for the image and encounter a problem other than image-hasn't-been-pulled (attempting to use an argument in the name would often lead to invalid-image-reference errors), we would previously have failed, but instead, we now let the build stage handle attempting to pull the image.

### DIFF
--- a/pkg/build/builder/common.go
+++ b/pkg/build/builder/common.go
@@ -392,6 +392,21 @@ func addBuildParameters(dir string, build *buildapiv1.Build, sourceInfo *git.Sou
 		return err
 	}
 
+	// Arguments set before the first stage get discarded when we split the
+	// tree up into stages in replaceImagesFromSource.  Other instructions
+	// are preserved.  Preserve the arguments here so that we can prepend
+	// them to the replacement Dockerfile that we generate.
+	var preambleArgs []byte
+	for _, child := range node.Children {
+		if child.Value == dockercmd.Arg {
+			preambleArgs = append(preambleArgs, []byte(child.Original+"\n")...)
+		} else {
+			if child.Value == dockercmd.From {
+				break
+			}
+		}
+	}
+
 	// Update base image if build strategy specifies the From field.
 	if build.Spec.Strategy.DockerStrategy != nil && build.Spec.Strategy.DockerStrategy.From != nil && build.Spec.Strategy.DockerStrategy.From.Kind == "DockerImage" {
 		// Reduce the name to a minimal canonical form for the daemon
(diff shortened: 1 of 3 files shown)


pkg/build/builder/common.go
```diff
@@ -425,19 +440,23 @@ func addBuildParameters(dir string, build *buildapiv1.Build, sourceInfo *git.Sou
 		return err
 	}
 
-	if err := replaceImagesFromSource(node, build.Spec.Source.Images); err != nil {
+	var buildArgs []corev1.EnvVar
```
gabemontero: Looks like you discovered a separate issue from the `ARG` one where docker strategy ENVs were getting lost as well @nalind 

I'm I reading that correctly?

If so, would a comment, analogous to your ARG explanation above, placed here, and a tweak to the PR title/commit, seem prudent?

gabemontero: And some new unit tests as well, again assuming I'm reading this correctly.

nalind: They're still build args.  The API struct is just called an `EnvVar`, I'm guessing because inventing a new type for every key-value pairing would have been excessive.

I'm not sure this added logic buys us anything, since we don't do any substitutions when we're walking the parsed tree, but I added it for completeness.  I wouldn't object if we wanted to drop it.

gabemontero: Thanks for the clarification.

I don't feel either way about keeping it at the moment, though perhaps a unit test would prove one way or the other whether it buys us anything.

Unless other chime in with stronger opinions, I'll leave it at that.

danielkucera:
Can we please get some attention here, this bug is affecting us, see: https://access.redhat.com/support/cases/#/case/02630699
if you have access.

gabemontero:
is there a bug associated with that case @danielkucera ? ... it would need to be associated with this PR in order for it to merge

tests are green

and my review comment was addressed at least

/assign @adambkaplan

danielkucera:
@gabemontero this was the original issue:
https://bugzilla.redhat.com/show_bug.cgi?id=1839683
but after latest update the error has changed to:
```
Caching blobs under "/var/cache/blobs".
error: build error: error parsing name "***/rhel8-ubi-base:${BASE_IMAGE_TAG}": error parsing image name "***/rhel8-ubi-base:${BASE_IMAGE_TAG}": invalid reference format
```
(registry was starred-out by me)

so now it depends on one of yours:
https://bugzilla.redhat.com/show_bug.cgi?id=1801875

gabemontero:
> @gabemontero this was the original issue:
> https://bugzilla.redhat.com/show_bug.cgi?id=1839683
> but after latest update the error has changed to:
> 
> ```
> Caching blobs under "/var/cache/blobs".
> error: build error: error parsing name "***/rhel8-ubi-base:${BASE_IMAGE_TAG}": error parsing image name "***/rhel8-ubi-base:${BASE_IMAGE_TAG}": invalid reference format
> ```
> 
> (registry was starred-out by me)
> 
> so now it depends on one of yours:
> https://bugzilla.redhat.com/show_bug.cgi?id=1801875

I see ... thanks for the info @danielkucera  .... I think we need to let @nalind make the call on which BZ he thinks should be associated with this PR.

After that is done, we'll need an approve from @adambkaplan and then either he or I to lgtm this

adambkaplan:
@nalind @gabemontero based on the discussion in this PR, I'm going to create a separate BZ to cover this specific case. There seem to be separate issues wrt args in Buildah that may need to be addressed.

adambkaplan:
/retitle 
Bug 1842982: preserve ARG values that come before the first FROM

adambkaplan:
/retitle Bug 1842982: preserve ARG values that come before the first FROM

PR #134, opened 2020-02-17
manage-dockerfile: use the original form of ENV/LABEL

nalind:
When rewriting the Dockerfile with our substitutions, we unconditionally insert `=` characters, even in cases where the original file didn't use them.  They're syntactically important, though, so such changes can cause builds to fail.  For ENV and LABEL instructions, just emit the original text.

Fixes https://github.com/openshift/origin/issues/24151.


nalind:
Added unit tests.

gabemontero:
Fixing through simplification

LGTM

adambkaplan:
/lgtm

@gabemontero already gave it.

PR #141, opened 2020-03-06
Bug 1810174: bump github.com/containers/buildah and dependencies

nalind:
Bump github.com/containers/buildah to v1.14.2, github.com/containers/image/v5 to v5.2.1, github.com/containers/storage to v1.16.1, and their various dependencies, largely to pull in https://github.com/containers/buildah/pull/2095.

Update for API changes in buildah:
* buildah.BuilderOptions no longer includes AddCapabilities and DropCapabilities slices, but now expects a unified Capabilities field which should be computed by passing those slices to github.com/containers/common/pkg/config.Default().Capabilities().
* imagebuildah.BuildOptions.TransientMounts is now a slice of string specifications for transient mounts rather than the structures we generate by parsing those specifications.


pkg/build/builder/daemonless.go
```diff
@@ -130,18 +130,13 @@ func buildDaemonlessImage(sc types.SystemContext, store storage.Store, isolation
...
 		// anything that the runtime mounted from the node into our
 		// /run/secrets, without passing along API secrets that were
 		// mounted under it exclusively for the builder's use.
-		transientMounts = append(transientMounts, imagebuildah.Mount{
-			Source:      "/run/secrets",
-			Destination: "/run/secrets",
-			Type:        "bind",
-			Options:     []string{"ro", "nodev", "noexec", "nosuid"},
-		})
+		transientMounts = append(transientMounts, "/run/secrets:/run/secrets:ro,nodev,noexec,nosuid")
```
nalind: Reviewer note: this specification eventually gets turned into a recursive bind mount by the newer API.  Is that a breaking change, or should I just update the comment to reflect the new behavior?

pkg/build/builder/source.go
```diff
@@ -392,12 +393,18 @@ func extractSourceFromImage(ctx context.Context, dockerClient DockerClient, stor
...
-		FromImage:        image,
-		PullPolicy:       pullPolicy,
-		ReportWriter:     os.Stdout,
-		SystemContext:    &systemContext,
-		DropCapabilities: dropCapabilities(),
+		FromImage:     image,
+		PullPolicy:    pullPolicy,
+		ReportWriter:  os.Stdout,
+		SystemContext: &systemContext,
+		Capabilities:  capabilities,
```
rhatdan: I thought you wanted to revert this change in Buildah?

nalind: We reverted the change in the `imagebuildah` package, but the main `buildah` package still included them.  If we’re also going to change those back, then it’ll resolve the `transientMounts` question above, else we’ll have to adapt.

pkg/build/builder/daemonless.go
adambkaplan: On the surface this doesn't appear to be a breaking change. The build controller doesn't add any content here, and from what I can gather recursive bind mounting is a desired outcome. No?

pkg/build/builder/source.go
adambkaplan: > else we’ll have to adapt.

@nalind what does this entail?

pkg/build/builder/daemonless.go
nalind: If the build controller doesn't depend on that, then it's fine.  I'll remove the bit from the comment about it being non-recursive.

pkg/build/builder/source.go
nalind: If the API changes aren't being reverted, then we'll still need the commit that adjusts us to the new API.  Otherwise we can drop it and keep only the one that updates our dependencies list and the vendored copies of the deps.

adambkaplan: @nalind I guess for me it comes down to whether or not this new API will be permanent, or if it will subsequently get reveted in buildah v1.15. If the former we can keep it as is.

nalind: @rhatdan what are your intentions there?

rhatdan: I thought you wanted me to revert these changes from buildah to work with the previous API.

adambkaplan:
@nalind this PR needs a rebase. We also need to merge and verify this fix before #144 is OK to merge.

PR #142, opened 2020-03-19
Bug 1809861: leverage new failed build status reasons/messages

gabemontero:
/hold 
looking into openshift/origin failure_status.go hits as well for this one 

/assgin @bparees 

@openshift/openshift-team-developer-experience @nalind FYI

@adambkaplan we got some k8s 17.0 to 17.1 bumpage as part of pulling in a more recent openshift/api (my only actual change to go.mod ... go mod tidy/vendor did the rest).


gabemontero:
`failed to acquire lease: status 503 Service Unavailable, status code 503` on e2e-aws-build

will re submit this evening assuming ci is under less pressure

pkg/build/builder/source.go
```diff
@@ -109,6 +110,9 @@ func ManageDockerfile(dir string, build *buildapiv1.Build) error {
 			baseDir = filepath.Join(baseDir, build.Spec.Source.ContextDir)
 		}
 		if err := ioutil.WriteFile(filepath.Join(baseDir, "Dockerfile"), []byte(*dockerfileSource), 0660); err != nil {
+			build.Status.Phase = buildapiv1.BuildPhaseFailed
+			build.Status.Reason = buildapiv1.StatusReasonManageDockerfileFailed
+			build.Status.Message = builderutil.StatusMessageManageDockerfileFailed
```
bparees: you need to make the api call to update the build somewhere in this call stack, e.g.: https://github.com/openshift/builder/blob/adc71a7dbf07918033f3b06b75988a71fef5458f/pkg/build/builder/cmd/builder.go#L239-L240

the ManageDockerfile callstack doesn't appear to currentl do that, so just updating the build object won't propagate the change back to etcd.

pkg/build/builder/util/consts.go
```diff
@@ -24,6 +24,8 @@ const (
 	StatusMessagePushImageToRegistryFailed       = "Failed to push the image to the registry."
 	StatusMessagePullBuilderImageFailed          = "Failed pulling builder image."
 	StatusMessageFetchSourceFailed               = "Failed to fetch the input source."
+	StatusMessageFetchImageContentFailed         = "Failed to extract image content."
+	StatusMessageManageDockerfileFailed          = "Failed to prepare the dockerfile for the build."
```
bparees: i'm a little surprised we're not including the error text in these messages, but i guess they should show up in LogSnippet, and you seem to be following the precedence we've established of not doing it (maybe we felt the errors were too raw to include in the message text)

gabemontero: +1

pkg/build/builder/source.go
gabemontero: ah missed that thanks @bparees 

in the update I'm about to push, since `ManageDockerfile` is not associated with the `builderConfig` struct I had to slightly mutate the exact lines of code used, but it was not onerous.

gabemontero:
update pushed @bparees

gabemontero:
non-build flakes e2e-aws

/test e2e-aws

still holding until get we get the test run we want in https://github.com/openshift/origin/pull/24731

gabemontero:
test failed as expected in https://github.com/openshift/origin/pull/24731

/hold cancel

PR #144, opened 2020-03-20
Bug 1817175: bump github.com/containers/buildah and dependencies

TomSweeneyRedHat:
Signed-off-by: TomSweeneyRedHat <tsweeney@redhat.com>

Emulates #141 but for the OCP 4.4 branch.  Stealing from @nalind 's PR notes from there:

Bump github.com/containers/buildah to v1.14.2, github.com/containers/image/v5 to v5.2.1, github.com/containers/storage to v1.16.1, and their various dependencies, largely to pull in containers/buildah#2095.

Update for API changes in buildah:

    buildah.BuilderOptions no longer includes AddCapabilities and DropCapabilities slices, but now expects a unified Capabilities field which should be computed by passing those slices to github.com/containers/common/pkg/config.Default().Capabilities().
    imagebuildah.BuildOptions.TransientMounts is now a slice of string specifications for transient mounts rather than the structures we generate by parsing those specifications.

PR #150, opened 2020-04-10
Bug 1817175: bump github.com/containers/buildah and dependencies

TomSweeneyRedHat:
Signed-off-by: TomSweeneyRedHat <tsweeney@redhat.com>

Signed-off-by: TomSweeneyRedHat <tsweeney@redhat.com>

Emulates #141 and #144 but for the OCP 4.3 branch.  @adambkaplan please let me know if this is not the right landing spot for 4.3.z.

Stealing from @nalind 's PR notes from there:

Bump github.com/containers/buildah to v1.14.2, github.com/containers/image/v5 to v5.2.1, github.com/containers/storage to v1.16.1, and their various dependencies, largely to pull in containers/buildah#2095.

Update for API changes in buildah:

    buildah.BuilderOptions no longer includes AddCapabilities and DropCapabilities slices, but now expects a unified Capabilities field which should be computed by passing those slices to github.com/containers/common/pkg/config.Default().Capabilities().
    imagebuildah.BuildOptions.TransientMounts is now a slice of string specifications for transient mounts rather than the structures we generate by parsing those specifications.


go.mod
```diff
@@ -8,24 +8,24 @@ require (
 	github.com/Nvveen/Gotty v0.0.0-20120604004816-cd527374f1e5 // indirect
 	github.com/armon/go-radix v1.0.0 // indirect
 	github.com/certifi/gocertifi v0.0.0-20180905225744-ee1a9a0726d2 // indirect
-	github.com/containers/buildah v1.11.6-0.20200109175634-4e23b7a10cc9
-	github.com/containers/image/v5 v5.1.0
-	github.com/containers/storage v1.15.5
+	github.com/containers/buildah v1.14.2
+	github.com/containers/common v0.4.2
+	github.com/containers/image/v5 v5.2.1
+	github.com/containers/storage v1.16.1
```
adambkaplan, line 14: @nalind what risks do we introduce by bumping buildah from 1.11.6 to 1.14.2? This is something that we haven't typically done in a z-stream release.

pkg/build/builder/source.go
```diff
@@ -392,12 +393,18 @@ func extractSourceFromImage(ctx context.Context, dockerClient DockerClient, stor
 		}
 	}
 
+	defaultContainerConfig, err := cconfig.Default()
+	if err != nil {
+		return err
+	}
```
adambkaplan, line 399: Does this have a high risk of erroring out here? Would we be able to easily identify errors from here?

go.mod
nalind: Hmm, then we might be better off trying to backport the desired change to buildah's `release-1.11-rhel` branch and switching to vendoring that for the `release-4.3` branch.

pkg/build/builder/source.go
nalind: After a quick spin through the code, it looks like we get enough detail in error results that we can usually pin down where an error occurred, if one does.
Missing `containers.conf` and `libpod.conf` files shouldn't create errors, though if we're unprivileged and we can't determine where the file should be, we get an error, but we're privileged here.
I think we're pretty well insulated from having new surprise errors if we pull this in.

adambkaplan:
/hold

Per https://github.com/openshift/builder/pull/150#discussion_r407701358, we should probably wait until the buildah fix is backported to the 1.11 release branch.

PR #155, opened 2020-06-02
Clean up imports

coreydaley:
Re-organizing the go imports


coreydaley:
/retest
failed to aquire lease

PR #158, opened 2020-06-18
Bug 1826183: Generate trust bundle for builds

adambkaplan:
* Create entrypoint script which does the following:
  * Check if the CA trust bundle exists in the new neutral location.
  * Run `update-ca-trust extract` if a custom PKI is present.
* Add mounts.conf to /etc/containers
  * Mount /run/secrets from build pod containers to buildah's containers.
   In RHEL/Fedora, /run/secrets is mounted in from /usr/share/rhel/secrets, and contains host
   information needed to access subscription content.
  * Mount /etc/pki from the build pod to buildah's containers. This contains the TLS trust store
   that the entrypoint script configures via `update-ca-trust extract`.
* Organize image content to simplify the Dockerfile instructions for building `openshift/builder`.


gabemontero:
/hold

the client plugin test is broke again via the upstream pipeline example in that repo @adambkaplan @akram @waveywaves 

I'm talking the file https://github.com/openshift/jenkins-client-plugin/blob/master/examples/jenkins-image-sample.groovy

I saw the same thing in recent samples repo PRs

I'll create a PR shortly that will move the build suite off of its use and instead onto a more stable pipeline in openshift/origin so we still have regression coverage

I am also working on  PR for jenkins-client-plugin that will finally allow e2e's in that repo to test against version of that test in that repo (and finally avoid the manual missteps that can occur testing that client plugin changes against the master branch vs. the version of the file in a given PR).

adambkaplan:
/hold

Placing a second hold on this based on the demo. The current chain of PRs will merge the trust bundle with the OS defaults. I need to check with the network team if this is acceptable.

/cc @danehans

danehans:
> The current chain of PRs will merge the trust bundle with the OS defaults.

@adambkaplan this is what the Network Operator does when an operator or operand requests proxy trust bundle injection, for [example](https://github.com/openshift/cluster-image-registry-operator/blob/master/manifests/04-ca-trusted.yaml). If you're not requesting trust bundle injection, then yes, you need to combine the OS default bundle with the bundle referenced by `proxy.spec.trustedCA`, mount the combined bundle and run `update-ca-trust extract`.

adambkaplan:
@danehans then it isn't clear if we are doing the right thing.

The openshift controller manager operator creates a ConfigMap for the injected CA, which is then read in by the build controller. When a build is started, the build controller creates a new ConfigMap with this CA data, and mounts the contents into the build pod. This was originally mounted into `/etc/pki/ca-trust/extracted`. https://github.com/openshift/openshift-controller-manager/pull/119/ changes the mount point, forcing us to copy the contents and run `update-ca-trust extract`.

Trust bundle injection is not sufficient because not all processes can use the TLS trust bundle. Java processes need the trust bundle in the keytool format.

danehans:
Plumbing the trust bundle configmap is up to the operator/operand implementation. For example, ingress operator mounts the trust bundle configmap to `/etc/pki/ca-trust/extracted/pem`. A file watcher was implemented that restarts the operator if a change is observed to the trust bundle file. Since ingress controllers (i.e. operand) does not need proxy support, the trusted ca bundle configmap does not need to get plumbed into any controllers.

adambkaplan:
/hold cancel

This PR can merge because it will run `update-ca-trust extract` with whatever trust bundle is provided.
Per @danehans comment, what we need to do is change the ConfigMap that the openshift controller manager receives from its operator. Instead of injecting the cluster CA, we should take the user CA directly from the openshift-config namespace and wire it through to build pods.

adambkaplan:
@gabemontero @coreydaley @otaviof ptal - I had to change the mount point for `/etc/pki/ca-trust`

coreydaley:
Is there any testing that goes along with this or is the existing case covered and will automatically pass if this change is done correctly?

adambkaplan:
@coreydaley https://github.com/openshift/origin/pull/25156 will verify once everything merges.

PR #126, opened 2020-01-10
Bug 1788982: bump github.com/containers/{storage,image,buildah}

nalind:
Update github.com/containers/storage from v1.15.3 to v1.15.5.
Update github.com/containers/image/v5 from v5.0.0 to v5.1.0.
Update github.com/containers/buildah to a newer snapshot, from a version slightly older than v1.12.0.

This should help with https://bugzilla.redhat.com/show_bug.cgi?id=1788982, though we'll need a separate PR for the release-4.3 branch.

PR #164, opened 2020-07-17
Bump containers/storage to v1.20.2

otaviof:
Updating `containers/storage` to `v1.20.2` in order to support using node's layers mounted on build pods. As per [PR #126 in `openshift/openshift-controller-manager`](https://github.com/openshift/openshift-controller-manager/pull/126).

Bumping `containers/storage` needs newer `buildah`, which are currently in `master`, plus minor refactoring on capturing error from `Capabilities` parsing.


otaviof:
@adambkaplan, @gabemontero, @nalind, please consider.

pkg/build/builder/source.go
```diff
@@ -434,7 +434,10 @@ func extractSourceFromImage(ctx context.Context, dockerClient DockerClient, stor
 	if err != nil {
 		return err
 	}
-	capabilities := defaultContainerConfig.Capabilities("", nil, dropCapabilities())
+	capabilities, err := defaultContainerConfig.Capabilities("", nil, dropCapabilities())
+	if err != nil {
+		return err
+	}
```
adambkaplan, line 440: @otaviof this come from the buildah bump?

go.mod
```diff
@@ -4,27 +4,29 @@ go 1.13
...
 require (
 	github.com/MakeNowJust/heredoc v0.0.0-20170808103936-bb23615498cd
-	github.com/containers/buildah v1.14.10
-	github.com/containers/common v0.8.4
-	github.com/containers/image/v5 v5.4.3
-	github.com/containers/storage v1.18.2
+	github.com/containers/buildah v1.15.1-0.20200716185836-11db2c1af990
+	github.com/containers/common v0.15.2
+	github.com/containers/image/v5 v5.5.1
+	github.com/containers/storage v1.20.2
```
adambkaplan, line 10: Not necessarily a _bad_ idea to bump buildah for a strict proof of concept. Personally I would have just run
```
go get github.com/containers/storage@v1.20.2
go mod tidy
go mod vendor
```
and left the `go mod` changes in place.

otaviof: That was my first try, @adambkaplan . However, it needs a bump in `buildah` as well, driven by the following error:
```
vendor/github.com/containers/buildah/run_linux.go:142:72: cannot use d.DeviceRule.Permissions (type configs.DevicePermissions) as type string in argument to g.AddLinuxResourcesDevice
make: *** [Makefile:11: build] Error 2
```

pkg/build/builder/source.go
otaviof: Yes, that's correct. And the reason for `buildah` bump explained below.

go.mod
nalind: That's from github.com/opencontainers/runtime-tools.  I'd strongly prefer we stay on the buildah release-1.15 branch if we can (11db2c1af990 is from the master branch), until we have a 1.16.

otaviof: @adambkaplan going back on this commit, I was able to do exactly what you mention. Please consider latest changes.

go.sum
```diff
@@ -722,7 +738,7 @@ github.com/openshift/library-go v0.0.0-20200327125526-163b2f0d6264 h1:VcZjTupwjx
 github.com/openshift/library-go v0.0.0-20200327125526-163b2f0d6264/go.mod h1:CfydoH0B+RYs22uQZQ36A1mz5m5zhucpMGh8t5s71v4=
 github.com/openshift/moby-moby v1.4.2-0.20190308215630-da810a85109d h1:1LuQzDKgiXj1omPNDcY1E/mEOE/90jdobR+7WBfBQYA=
 github.com/openshift/moby-moby v1.4.2-0.20190308215630-da810a85109d/go.mod h1:LJM49W8fBVSj+rvcopJZu9mgH5Tx6HwLHySIYeGeu4k=
-github.com/openshift/source-to-image v1.3.0 h1:YDEH0PY0sMaUNbpJAGnDZA94tkAD6qi1Rou8ycWRPRc=
+github.com/openshift/source-to-image v1.3.0 h1:ZktIgJ85Vn0HOyHdS6uQOMXhmYwc5h76w/6LpnJ60L8=
```
otaviof, line 741: @adambkaplan that's a strange behavior on working with `go mod` for this PR, I had to remove `source-to-image` line from `go.sum`, in order to match the correct commit checksum.

rhatdan:
/approve
LGTM
@nalind PTAL

rhatdan:
@adambkaplan PTAL

go.sum
adambkaplan: think this is because I merged a few doc changes into s2i and moved the tag. No golang code changed.

vendor/modules.txt
```diff
@@ -47,7 +48,7 @@ github.com/blang/semver
 github.com/certifi/gocertifi
 # github.com/containerd/cgroups v0.0.0-20190919134610-bf292b21730f
 github.com/containerd/cgroups/stats/v1
-# github.com/containerd/containerd v1.3.0 => github.com/containerd/containerd v1.3.0
+# github.com/containerd/containerd v1.3.2 => github.com/containerd/containerd v1.3.0
```
adambkaplan, line 51: @otaviof so it looks like a transitive dep now uses `containerd/containerd v1.3.2`. We have a replace statement to pin containerd to v1.3.0 - please update the replace to use the new tag.

go.sum
otaviof: Thanks.

vendor/modules.txt
otaviof: Just pushed the changes to make sure `containerd/containerd` version `1.3.2` in place. Please consider.

PR #165, opened 2020-07-22
Add Dockerfile.rhel8

yselkowitz:
A separate file is required, as the changes necessary for building with
RHEL 8 content are not compatible with RHEL 7.

### DIFF
--- a/Dockerfile.rhel8
+++ b/Dockerfile.rhel8
@@ -1,0 +1,30 @@
+FROM registry.svc.ci.openshift.org/ocp/builder:golang-1.13 AS builder
+WORKDIR /go/src/github.com/openshift/builder
+COPY . .
+RUN hack/build.sh
+
+FROM registry.svc.ci.openshift.org/ocp/4.0:base-8
+RUN INSTALL_PKGS=" \
+      bind-utils bsdtar findutils fuse-overlayfs git hostname lsof \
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM
/test e2e-aws

Dockerfile.rhel8
```diff
@@ -0,0 +1,30 @@
+FROM registry.svc.ci.openshift.org/ocp/builder:golang-1.13 AS builder
+WORKDIR /go/src/github.com/openshift/builder
+COPY . .
+RUN hack/build.sh
+
+FROM registry.svc.ci.openshift.org/ocp/4.0:base
```
adambkaplan, line 6: Is this currently a ubi-8 based image, or @yselkowitz are you using a `BuildConfig` or OpenShift CI to substitute this with the UBI8 image?

If we want to run this Dockerfile "as is", perhaps this should be replaced with:
```suggestion
FROM registry.access.redhat.com/ubi8/ubi:latest
```

yselkowitz: Sorry, my focus has been on CI and ART substitution, where the FROMs are replaced anyway.  iiuc 4.0:base is a placeholder anyway, but it should probably be base-8 instead.  WIll fix accordingly.

PR #181, opened 2020-10-21
fixing version string generation to not be bound to env vars

coreydaley-redhat:


adambkaplan:
/hold

I don't quite know how this impacts version numbing in OSBS. Are there alternatives in openshift/build-machinery-go that we can use?

coreydaley-redhat:
/close

Just need to run "git fetch upstream" and the issue goes away.

PR #189, opened 2020-11-18
Bug 1896446: Fix private git clones behind a proxy 

adambkaplan:
Switch between the "default" and global git configuration depending if a
.gitconfig file has been generated:

- Use the "default" .gitconfig specified by the GIT_CONFIG environment variable
  if a git configuration file was generated when source secrets were used to
  initialize the git envrionment.
- Use the global .gitconfig if no source secret was set up.
- Updated godoc to indicate when a .gitconfig file was generated and return values
  for setupGitEnvironment.
- Add missing methods to the SCMAuthContext interface. Update functions
to return SCMAuthContext interfaces.
- Add test package for unit test framework code


gabemontero:
oooh if this works with existing e2e's and the source secret / proxy combo I think I like this better than mine @adambkaplan

adambkaplan:
@gabemontero throwing this up to see what happens. This switches us from using the `--global` .gitconfig to the "local" one that sits in the `secretTmpDir`. I _think_ is always set (it configured via the `SOURCE_SECRET_PATH` env var), but am not sure.

gabemontero:
> @gabemontero throwing this up to see what happens. This switches us from using the `--global` .gitconfig to the "local" one that sits in the `secretTmpDir`. I _think_ is always set (it configured via the `SOURCE_SECRET_PATH` env var), but am not sure.

wait ... I think then this will fix the proxy + source secret case, but not the 
proxy without source secret case

yeah I would expect a form in `setupProxyConfig` that uses `--global` when no source secret exists, but does your new add local when it does exist.

adambkaplan:
> wait ... I think then this will fix the proxy + source secret case, but not the
> proxy without source secret case
> 
> yeah I would expect a form in `setupProxyConfig` that uses `--global` when no source secret exists, but does your new add local when it does exist.

You are right - all the existing proxy tests fall over.

gabemontero:
relevant e2e's clean @adambkaplan 

e2e-aws failures are unrelated flakes

adambkaplan (commented):
/assign @coreydaley 

This was the PR @gabemontero mentioned in scrum to address the git clone secret + proxy issue.

pkg/build/builder/cmd/builder.go
```diff
@@ -226,30 +226,41 @@ func (c *builderConfig) setupGitEnvironment() (string, []string, error) {
...
 	if gitSource.HTTPProxy != nil && len(*gitSource.HTTPProxy) > 0 {
-		err := gitClient.AddGlobalConfig("http.proxy", *gitSource.HTTPProxy)
+		err := c.addGitConfig(gitClient, gitConfigDir, "http.proxy", *gitSource.HTTPProxy, useLocal)
 		if err != nil {
 			return err
 		}
 	}
 	if gitSource.HTTPSProxy != nil && len(*gitSource.HTTPSProxy) > 0 {
-		err := gitClient.AddGlobalConfig("https.proxy", *gitSource.HTTPSProxy)
+		err := c.addGitConfig(gitClient, gitConfigDir, "https.proxy", *gitSource.HTTPSProxy, useLocal)
```
coreydaley, line 249: if err := c.addGitConfig(gitClient, gitConfigDir, "https.proxy", *gitSource.HTTPSProxy, useLocal); err != nil {
			return err
		}

pkg/build/builder/cmd/builder.go
```diff
@@ -226,30 +226,41 @@ func (c *builderConfig) setupGitEnvironment() (string, []string, error) {
...
 		if err != nil {
 			return err
 		}
 	}
 	return nil
 }
 
+func (c *builderConfig) addGitConfig(gitClient git.Repository, gitConfigDir string, parameter string, value string, local bool) error {
```
coreydaley, line 257: s/local/useLocal/
I feel like local is to ambiguous, if it's a bool it should be something like "isLocal" or "useLocal" depending on it's usage and meaning.

pkg/build/builder/cmd/builder.go
```diff
@@ -265,7 +276,7 @@ func (c *builderConfig) clone() error {
 	defer os.RemoveAll(secretTmpDir)
 
 	gitClient := git.NewRepositoryWithEnv(gitEnv)
-	err = c.setupProxyConfig(gitClient)
+	err = c.setupProxyConfig(gitClient, secretTmpDir)
```
coreydaley, line 279: if err = c.setupProxyConfig(gitClient, secretTmpDir); err != nil {
		return err
	}

pkg/build/builder/cmd/builder_test.go
```diff
@@ -0,0 +1,266 @@
...
+					},
+				},
+			},
+			expectedGlobalConfigs: map[string]string{
+				"http.proxy":  fakeHTTPProxy,
+				"https.proxy": fakeHTTPSProxy,
+			},
+		},
+		{
+			name: "git clone proxy",
```
coreydaley, line 77: s/git clone proxy/git clone auth no proxy/

pkg/build/builder/cmd/builder_test.go
```diff
@@ -0,0 +1,266 @@
...
+				if actual != expected {
+					t.Errorf("expected global git config %s to be %s, got %s", config, expected, actual)
+				}
+			}
+		})
+	}
+
+}
+
+func newFakeGitRepository() *fakeGitRepository {
```
coreydaley, line 173: nit: This and below should possibly be factored out into a test framework like we had for the ciro https://github.com/openshift/cluster-image-registry-operator/tree/master/test/framework

coreydaley (changes requested):

adambkaplan:
/bugzilla cc-qa

We've had a tough time spinning up a ci-bot cluster with proxy. If we get this verified in 4.7 we'd like to backport 4.6.z ASAP (currently blocking promotion to stable)

pkg/build/builder/cmd/builder_test.go
```diff
@@ -0,0 +1,266 @@
...
+								},
+							},
+							SourceSecret: &corev1.LocalObjectReference{
+								Name: "git-auth",
+							},
+						},
+					},
+				},
+			},
+			expectedLocalConfigs: map[string]string{
```
gabemontero, line 123: so I see us validating the proxy stuff is there, but maybe based on the 2 source secret types we support for the proxy scenario (userid pwd and included .gitconfig) we should verify that their content was not removed ... i.e. we are in fact updating the same file @adambkaplan 

so if we 
a) leverage the k8s fake secret client stuff for tests
b) create a secret that looked like 
    i) the basic auth style of secret fashioned from https://docs.openshift.com/container-platform/4.6/builds/creating-build-inputs.html#builds-source-secret-basic-auth_creating-build-inputs
   ii) the include your own .gitconfig file secret per https://docs.openshift.com/container-platform/4.6/builds/creating-build-inputs.html#builds-gitconfig-file-secured-git_creating-build-inputs

we would fully could fully validate that the new git use local config stuff updates / includes that stuff accordingly

In particular, consider the password case where it creates `gitcredentials.*` file to include in the "git config" file it names `gitcredentialscfg.`

let's make sure after your local call the same file has everything.

the user provided .gitconfig file case is probably more straight forward

WDYT ?

pkg/build/builder/cmd/builder_test.go
adambkaplan: updated - moved to a "fake" package since it's used for unit testing.

pkg/build/builder/cmd/builder.go
```diff
@@ -188,29 +188,31 @@ func newBuilderConfigFromEnvironment(out io.Writer, needsDocker bool) (*builderC
...
 		sourceURL, err := s2igit.Parse(gitSource.URI)
 		if err != nil {
-			return "", nil, fmt.Errorf("cannot parse build URL: %s", gitSource.URI)
+			return "", nil, "", fmt.Errorf("cannot parse build URL: %s", gitSource.URI)
 		}
 		scmAuths := scmauth.GitAuths(sourceURL)
 
-		secretsEnv, overrideURL, err := scmAuths.Setup(c.sourceSecretDir)
+		secretsEnv, overrideURL, gitConfigFile, err := scmAuths.Setup(c.sourceSecretDir)
+		outGitConfigFile = gitConfigFile
```
coreydaley, line 213: I don't see why you are doing this, I can not seem to find anywhere else that you are using gitConfigFile anywhere in this function.

coreydaley:
I feel like the "fake" package should be in the test directory, as it should not be held to the same standard as production code.

pkg/build/builder/cmd/builder.go
adambkaplan: This is passed through to `setupProxyConfig` in line 272. I'll add a godoc to make the returns clearer.

coreydaley: Isn't that in a different function? Why are you copying the variable?

adambkaplan:
/bugzilla cc-qa

I was able to do some testing on @xiuwang's proxy cluster, which only allows access to the Internet via the proxy. I created a build using a private git repo of mine and the following combination of creds and NO_PROXY settings:

1. Basic HTTP auth with token. Success.
2. Basic HTTP auth with token, github.com added to NO_PROXY. Times out cloning source (cannot access github.com)
3. Basic HTTP auth without token. Fails cloning source (auth error)
4. Basic HTTP auth without token, github.com added to NO_PROXY. Times out failing to clone source (cannot access github.com)
5. SSH clone with private key. Times out cloning source (cannot access github.com)
6. SSH clone without private key. Times out cloning source (cannot access github.com)

Bottom line is I have a higher degree of confidence this will pass verification. NO_PROXY is being respected for HTTP/S cloning. SSH is failing because we don't support ssh-based clone behind a proxy.

pkg/build/builder/cmd/builder_test.go
adambkaplan: IMO we don't have to go so far as to wire in a fake openshift client. We can probably cover the items you are talking about with a unit test for `setupGitEnvironment`. I'll include one to make sure that it is returning a path to a `.gitconfig` which has appropriate include clauses.

adambkaplan: I tried writing a good unit test for this function. Unfortunately there are side-effects which write files to temp dirs that are difficult to track. In general I don't like adding tests which can't clean themselves up.

Added https://issues.redhat.com/browse/BUILD-194 to address.

cc @coreydaley @otaviof

xiuwang:
Launched a 4.7 proxy cluster, using image built from 'build openshift/builder#189'.
Test scenarios:
1. With .gitconfig of source secret  , build could pull code from private repo. Builds go to completed.
2. With basic auth of source secret, build could pull code from private repo. Builds go to completed.
3. Basic HTTP auth with token, github.com added to NO_PROXY. Times out cloning source (cannot access github.com)

xiuwang:
/lgtm
Test passed on the 4.7 proxy cluster using image built from 'build #189'.
With .gitconfig of source secret , build could pull code from private repo. Builds go to completed.
With basic auth of source secret, build could pull code from private repo. Builds go to completed.

PR #192, opened 2020-11-25
[release-4.6] Bug 1901512: Fix private git clones behind a proxy

adambkaplan:
Manual cherry-pick of #189, due to library-go bump.


xiuwang:
Test on proxy cluster built from openshift/builder#192.

Test source secret with .gitconfig file , build could pull code from private repo. Builds go to completed.
Test source secret with basic auth, build could pull code from private repo. Builds go to completed.
Basic HTTP auth, github.com added to NO_PROXY. Times out cloning source (cannot access github.com)

PR #200, opened 2021-01-08
Bug 1918879: better surface dockerconfigjson errors causing image pull errors

gabemontero:
As part of the triage captured at https://coreos.slack.com/archives/C01JLQ7MB2M for a recent CI incident where a corrupted pull secret caused image pull failures with builds, it became apparent that V(4) trace was minimally needed to more precisely capture the error.

The PR strives to make the error cause more available with default logging.

@stevekuznetsov @alvaroaleman FYI

/assign @adambkaplan 

@openshift/openshift-team-build-api FYI

### DIFF
--- a/pkg/build/builder/cmd/dockercfg/cfg.go
+++ b/pkg/build/builder/cmd/dockercfg/cfg.go
@@ -10,6 +10,7 @@ import (
 	docker "github.com/fsouza/go-dockerclient"
 	"github.com/spf13/pflag"
 
+	kerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/klog/v2"
 	"k8s.io/kubernetes/pkg/credentialprovider"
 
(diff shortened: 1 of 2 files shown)


stevekuznetsov (commented):
WIll this error message leak token content?

pkg/build/builder/cmd/dockercfg/cfg.go
```diff
@@ -116,46 +118,53 @@ func GetDockercfgFile(path string) string {
...
 	return readDockerConfigJSONFileFromBytes(contents)
 }
 
-func readDockerConfigJSONFileFromBytes(contents []byte) bool {
+func readDockerConfigJSONFileFromBytes(contents []byte) error {
 	var cfgJSON credentialprovider.DockerConfigJSON
 	if err := json.Unmarshal(contents, &cfgJSON); err != nil {
-		log.V(4).Infof("while trying to parse blob %q: %v", contents, err)
-		return false
+		log.V(0).Infof("error trying to parse blob %q: %v", contents, err)
```
alvaroaleman, line 135: Yeah, yesterday it looked like it ends up printing part of the secret, I don't think we can print the error itself by default. What about just printing "Error Deserializing, set env var X to value Y to see details (warning: This might leak the content of the secret"?

gabemontero: yep I'll redo this part

gabemontero: and to be clear, it was the `contents` piece here ... if we considered the error log that occurred here, `while trying to parse blob << the raw json pull secret >> ": illegal base64 data at input byte 1152`

The `contents` piece is `<< the raw json pull secret >>` and the err string is `illegal base64 data at input byte 1152`

I'll leave the err string

gabemontero: I've also double checked `json.Unmarshal`  ... all the errors just note the byte array location like in the log above

gabemontero:
@stevekuznetsov @alvaroaleman separate commit pushed to remove byte array contents from the error log

pkg/build/builder/cmd/dockercfg/cfg.go
```diff
@@ -116,46 +118,53 @@ func GetDockercfgFile(path string) string {
...
 				log.V(4).Infof("found valid %s at %s", file, absDockerConfigFileLocation)
-				return absDockerConfigFileLocation
+				return absDockerConfigFileLocation, nil
 			}
+			errList = append(errList, err)
 		}
 	}
-	return ""
+	var err error
+	if len(errList) > 0 {
```
alvaroaleman, line 164: Nit: The condition isn't needed, utilerrors.NewAggregate will return a nil error if the error list has zero entries or all of them are nil

gabemontero: ah right thanks ... update pushed in separate commit

alvaroaleman (approved):
/lgtm

gabemontero:
thanks @stevekuznetsov @alvaroaleman review commits squashed awaiting look-see from @adambkaplan

adambkaplan (approved):
/approve

PR #202, opened 2021-01-13
Force use of runc, since we know we're privileged, and customize our seccomp filter

nalind:
* Force OCI isolation using runc instead of whatever we're configured for by the build controller (currently, usually chroot) to get control group settings that chroot isolation doesn't implement.
* Replace the default seccomp filter with one based on Fedora 33's, but with mount and umount disabled even for RUN instructions that run as root.

### DIFF
--- a/pkg/build/builder/daemonless.go
+++ b/pkg/build/builder/daemonless.go
@@ -259,6 +259,10 @@ func buildDaemonlessImage(sc types.SystemContext, store storage.Store, isolation
 		transientMounts = append(transientMounts, "/run/secrets:/run/secrets:ro,nodev,noexec,nosuid")
 	}
 
+	// Use a profile provided in the image instead of the default provided
+	// in runtime-tools's generator logic.
+	seccompProfilePath := "/usr/share/containers/seccomp.json"
+
 	options := imagebuildah.BuildOptions{
 		ContextDirectory: contextDir,
 		PullPolicy:       pullPolicy,
@@ -275,11 +279,12 @@ func buildDaemonlessImage(sc types.SystemContext, store storage.Store, isolation
 			{Name: string(specs.NetworkNamespace), Host: true},
 		},
 		CommonBuildOpts: &buildah.CommonBuildOptions{
-			HTTPProxy:    true,
-			Memory:       opts.Memory,
-			MemorySwap:   opts.Memswap,
-			CgroupParent: opts.CgroupParent,
-			Ulimit:       daemonlessProcessLimits(),
+			HTTPProxy:          true,
+			Memory:             opts.Memory,
+			MemorySwap:         opts.Memswap,
+			CgroupParent:       opts.CgroupParent,
+			Ulimit:             daemonlessProcessLimits(),
+			SeccompProfilePath: seccompProfilePath,
 		},
 		Layers:                  layers,
 		NoCache:                 opts.NoCache,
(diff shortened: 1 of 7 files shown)


gabemontero:
fwiw all the actual build related tests in https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_builder/202/pull-ci-openshift-builder-master-e2e-aws-builds/1349410553700814848 passed

at first blush the items that failed were unrelated to this PR

/retest

gabemontero:
the image-eco failure is a known item ... scl team sufficiently changed the usage message on their images to break on of our tests ... the tests need to be updated (their change seems fair to me)

gabemontero:
clean e2e-aws and e2e-aws-builds @nalind !

pkg/build/builder/daemonless.go
```diff
@@ -555,6 +561,9 @@ func GetDaemonlessClient(systemContext types.SystemContext, store storage.Store,
 		return nil, fmt.Errorf("unrecognized BUILD_ISOLATION setting %q", strings.ToLower(isolationSpec))
 	}
 
+	isolation = buildah.IsolationOCI
```
gabemontero, line 564: Nalin did this for convenience but you can also force this by setting the `BUILD_ISOLATION` env var on the build

gabemontero:
image eco should be fixed

/retest

/hold

pkg/build/builder/daemonless.go
gabemontero, line 564: Ok @nalind  I think we should tweak this slightly to where we supply a default of `IsolationOCI` but cans still override it vai the `BUILD_ISOLATION` env var (in case say there are performance implications for using `IsolationOCI` vs. `IsolationChroot`)

As such, the case statement above here at https://github.com/openshift/builder/blob/master/pkg/build/builder/daemonless.go#L555 should change the default case to simply pick OCI vs. returning an error

Then, after this merges.  We can craft an OCM PR where we remove the setting of `BUILD_ISOLATION` to chroot at https://github.com/openshift/openshift-controller-manager/blob/master/pkg/build/controller/strategy/util.go#L457 and https://github.com/openshift/openshift-controller-manager/blob/master/pkg/build/controller/strategy/util.go#L491

thoughts @nalind @bparees ?

pkg/build/builder/daemonless.go
```diff
@@ -259,6 +259,11 @@ func buildDaemonlessImage(sc types.SystemContext, store storage.Store, isolation
 		transientMounts = append(transientMounts, "/run/secrets:/run/secrets:ro,nodev,noexec,nosuid")
 	}
 
+	seccompProfilePath := "/usr/share/containers/seccomp.json"
+	if profileEnv, ok := os.LookupEnv("BUILD_SECCOMP_PROFILE"); ok {
+		seccompProfilePath = profileEnv
+	}
+
```
gabemontero, line 266: @nalind per my comment on how to default, would this change to add teh seccomp only be done if we were OCI isolation ?

gabemontero (changes requested):
I think we should go with defaulting but still being configurable ... see details below

gabemontero:
removing hold .... we'll sort out my comments via normal PR review 

/hold cancel

pkg/build/builder/daemonless.go
nalind: Having OCM change what it tells the builder image to use is the more technically correct way to do this, since we've been providing that particular control knob for some time.  Forcing it here was also about flushing out possible breakage for that case, but all things being equal I'd prefer to let OCM continue to make that decision.

pkg/build/builder/daemonless.go
nalind: The seccomp filter is also applied when using chroot isolation.  The magic value "unconfined" is recognized as instructing us to not add a filter.

gabemontero: turns out `BUILD_SECCOMP_PROFILE` is not an env var in the approved list at https://github.com/openshift/api/blob/master/build/v1/consts.go#L169

similar to the removing the case statement on the isolation, let remove the if check and just set seccompProfilePath to the default location.

gabemontero (changes requested):
One more change I believe is needed @nalind 

once you do that, go ahead and squash the commits and I'll assign to @bparees for final approval / merge (I'm not an approver in the OWNERS file so I cannot do it).

pkg/build/builder/daemonless.go
nalind: Changed it.

gabemontero:
/lgtm

/assign @bparees 

for adding approve

PR #203, opened 2021-01-13
[release-4.6] force use of runc, since we know we're privileged, and customize our seccomp filter

nalind:
* Force OCI isolation using runc instead of whatever we're configured for by the build controller (currently, usually chroot) to get control group settings that chroot isolation doesn't implement.
* Replace the default seccomp filter with one based on Fedora 33's, but with listen, accept, mount, and umount disabled even for RUN instructions that run as root.


nalind:
Closing until after we decide whether we want to do any of this on the master branch.

PR #206, opened 2021-01-19
Bug 1916897: narrow scope of rhsm transient bind mount

gabemontero:
/assign @bparees 

@nalind @smarterclayton

### DIFF
--- a/pkg/build/builder/daemonless.go
+++ b/pkg/build/builder/daemonless.go
@@ -32,6 +32,7 @@ import (
 
 	"github.com/openshift/builder/pkg/build/builder/cmd/dockercfg"
 	builderutil "github.com/openshift/builder/pkg/build/builder/util"
+	s2ifs "github.com/openshift/source-to-image/pkg/util/fs"
 )
 
 var (
@@ -259,10 +260,22 @@ func buildDaemonlessImage(sc types.SystemContext, store storage.Store, isolation
 	}
 
 	var transientMounts []string
-	if st, err := os.Stat("/run/secrets"); err == nil && st.IsDir() {
-		// Add a bind of /run/secrets, to pass along anything that the
-		// runtime mounted from the node into our /run/secrets.
-		transientMounts = append(transientMounts, "/run/secrets:/run/secrets:ro,nodev,noexec,nosuid")
+	if st, err := os.Stat("/run/secrets/rhsm"); err == nil && st.IsDir() {
+		// Add a bind of /run/secrets/rhsm, to pass along anything that the
+		// runtime mounted from the node into our /run/secrets/rhsm.
+		log.V(0).Infof("Adding transient rw bind mount for /run/secrets/rhsm")
+		tmpDir, err := ioutil.TempDir("/tmp", "rhsm-copy")
+		if err != nil {
+			log.V(0).Infof("Error creating tmpdir to set up /run/secrets/rhsm in build container: %s", err.Error())
+			return err
+		}
+		fs := s2ifs.NewFileSystem()
+		err = fs.CopyContents("/run/secrets/rhsm", tmpDir, map[string]string{})
+		if err != nil {
+			log.V(0).Infof("Error copying /run/secrets/rhsm to tmpdir %s: %s", tmpDir, err.Error())
+			return err
+		}
+		transientMounts = append(transientMounts, fmt.Sprintf("%s:/run/secrets/rhsm:rw,nodev,noexec,nosuid", tmpDir))
 	}
 
 	// Use a profile provided in the image instead of the default provided
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -2,8 +2,6 @@ package main
 
 import (
 	"fmt"
-	"io"
-	"io/ioutil"
 	"math/rand"
 	"os"
 	"os/signal"
@@ -17,10 +15,10 @@ import (
 
 	"k8s.io/component-base/logs"
 
-	"github.com/openshift/library-go/pkg/serviceability"
-
 	"github.com/openshift/builder/pkg/build/builder"
 	"github.com/openshift/builder/pkg/version"
+	"github.com/openshift/library-go/pkg/serviceability"
+	s2ifs "github.com/openshift/source-to-image/pkg/util/fs"
 )
 
 func main() {
@@ -51,14 +49,15 @@ func main() {
 
 	clusterCASrc := fmt.Sprintf("%s/ca.crt", builder.SecretCertsMountPath)
 	clusterCADst := fmt.Sprintf("%s/cluster.crt", tlsCertRoot)
-	err := CopyFileIfExists(clusterCASrc, clusterCADst)
+	fs := s2ifs.NewFileSystem()
+	err := fs.Copy(clusterCASrc, clusterCADst, map[string]string{})
 	if err != nil {
 		fmt.Printf("Error setting up cluster CA cert: %v", err)
 		os.Exit(1)
 	}
 
 	runtimeCASrc := fmt.Sprintf("%s/certs.d", builder.ConfigMapCertsMountPath)
-	err = CopyDirIfExists(runtimeCASrc, runtimeCertRoot)
+	err = fs.CopyContents(runtimeCASrc, runtimeCertRoot, map[string]string{})
 	if err != nil {
 		fmt.Printf("Error setting up service CA cert: %v", err)
 		os.Exit(1)
@@ -69,60 +68,6 @@ func main() {
 	if err := command.Execute(); err != nil {
 		os.Exit(1)
 	}
-}
-
-// CopyDirIfExists recursively copies a directory to the destination path.
-// If the source directory does not exist, no error is returned.
-// If the destination directory exists, any contents with matching file names
-// will be overwritten.
-func CopyDirIfExists(src, dst string) error {
-	srcInfo, err := os.Stat(src)
-	if os.IsNotExist(err) {
-		return nil
-	}
-	if err = os.MkdirAll(dst, srcInfo.Mode()); err != nil {
-		return err
-	}
-	dirInfo, err := ioutil.ReadDir(src)
-	for _, info := range dirInfo {
-		srcPath := filepath.Join(src, info.Name())
-		dstPath := filepath.Join(dst, info.Name())
-		if info.IsDir() {
-			err = CopyDirIfExists(srcPath, dstPath)
-		} else {
-			err = CopyFileIfExists(srcPath, dstPath)
-		}
-		if err != nil {
-			return err
-		}
-	}
-	return nil
-}
-
-// CopyFileIfExists copies the source file to the given destination, if the source file exists.
-// If the destination file exists, it will be overwritten and will not copy file attributes.
-func CopyFileIfExists(src, dst string) error {
-	_, err := os.Stat(src)
-	if os.IsNotExist(err) {
-		return nil
-	}
-	in, err := os.Open(src)
-	if err != nil {
-		return err
-	}
-	defer in.Close()
-
-	out, err := os.Create(dst)
-	if err != nil {
-		return err
-	}
-	defer out.Close()
-
-	_, err = io.Copy(out, in)
-	if err != nil {
-		return err
-	}
-	return out.Close()
 }
 
 // CommandFor returns the appropriate command for this base name,


nalind:
Would we get the same effect by removing the code block the patch is modifying, and adding the spec to `imagecontent/etc/containers/mounts.conf`?

pkg/build/builder/daemonless.go
```diff
@@ -253,10 +253,11 @@ func buildDaemonlessImage(sc types.SystemContext, store storage.Store, isolation
...
 	var transientMounts []string
-	if st, err := os.Stat("/run/secrets"); err == nil && st.IsDir() {
-		// Add a bind of /run/secrets, to pass along anything that the
-		// runtime mounted from the node into our /run/secrets.
-		transientMounts = append(transientMounts, "/run/secrets:/run/secrets:ro,nodev,noexec,nosuid")
+	if st, err := os.Stat("/run/secrets/rhsm"); err == nil && st.IsDir() {
+		// Add a bind of /run/secrets/rhsm, to pass along anything that the
+		// runtime mounted from the node into our /run/secrets/rhsm.
+		log.V(0).Infof("Adding transient rw bind mount for /run/secrets/rhsm")
+		transientMounts = append(transientMounts, "/run/secrets/rhsm:/run/secrets/rhsm:rw,nodev,noexec,nosuid")
```
bparees, line 260: @nalind with this approach will writes to /run/secrets/rhsm from within the build container propagate to /run/secrets/rhsm in the build pod?  (we do not want that)

nalind: The bind mount will let it modify content in the build container.  Unless I'm misreading things, the node's `/usr/share/containers/mounts.conf` includes `/usr/share/rhel/secrets:/run/secrets` on a 4.6.6 cluster, so cri-o creates a copy of the contents of `/usr/share/rhel/secrets` and bind mounts it for each container that it creates.  While modifying the `/run/secrets` in the build container would modify its copy of the content, each container in the build pod has its own copy.

bparees: i don't think we want to allow any actions in the build container to affect content in the build pod containers.

nalind: Right, the other containers in the pod have their own copies of that content.

gabemontero: OK we have all green tests with this current form.  Given the discussion above, and @nalind 's https://github.com/openshift/builder/pull/206#issuecomment-763070962 do we want to try a separate commit where I remove this piece of code and re-introduce a master branch version of https://github.com/openshift/builder/blob/release-4.6/imagecontent/etc/containers/mounts.conf where we change `/run/secrets` to `/run/secrets/rhsm` ?

I'm also curious on seeing the `ls -R /run` dump from the new test in https://github.com/openshift/origin/pull/25810 to make sure there is nothing else we need to consider

bparees: I think I am failing at terminology here.

We have:
1) the node filesystem
2) the pod container's filesystem (multiple of these, but i only care about the one that's running the buildah code)
3) the filesystem of the container/chroot/whatever that buildah created and in which user logic is being executed (the dockerfile commands).

A command that writes to something in (3) must not affect any container filesystem in (2).

Right now it sounds like a write to something in (3) (/run/secrets/rhsm) is going to propagate to the pod container that is running buildah, in (2).

gabemontero:
OK @nalind @bparees I've pushed in a separate commit the alternate approach of removing the code in daemonless.go and re-introducing the mounts.conf file, but using /run/secrets/rhsm

I'm still trying to get a cluster with these change to try the DPTP Dockerfile with the chmod's ( a couple of clusterbot attempts have failed)

pkg/build/builder/daemonless.go
nalind: Entries in the `transientMounts` slice are bind mounts to the container's filesystem, which I think is case (2).  Making them read-write would allow commands run by RUN to modify them, and those changes would be preserved for subsequent RUN instructions and commands run in the same container.

For entries in `mounts.conf`, a copy of the data in the source location is created in a temporary location, and the bind mount is set up using that copy as the source location.  This happens separately for each RUN instruction, which is more like (3) but with no persistence between even different RUN instructions that I think (3) expects.

imagecontent/etc/containers/mounts.conf
```diff
@@ -0,0 +1 @@
+/run/secrets/rhsm:/run/secrets/rhsm
```
bparees, line 1: i think my concern that this results in writing changes through to the build pod's container remains.

pkg/build/builder/daemonless.go
bparees: ooook.

so how do we get "creates a copy when it sets up the mount, uses that copy for all RUN commands, allows RUN commands to write to the mount"?  because that is what we want :)

bparees: (if we can't easily get that, short of doing the copy ourselves prior to invoking buildah, we can probably live with what mounts.conf gets us since i doubt people are writing to this dir and caring about it)

nalind: As you've guessed, persisting that across the build is not a feature that's built-in.  I've just noticed that it persists this way with `ImageOptimizationSkipLayers`, but I think that's a bug.

xiuwang:
Launched cluster from this cluster, and trigger build from DPTP bc http://file.rdu.redhat.com/~hongkliu/dptp2021/bz1916897/content-mirror/app.ci.bc.content-mirror.yaml.

Build content-mirror-2 go to complete without /run/secret read only error.
$ oc get builds -n ci
NAME               TYPE     FROM          STATUS     STARTED             DURATION
content-mirror-2   Docker   Git@083b29e   Complete   About an hour ago   1m16s

Build httpd-ex-2 with source secret go to complete too.
$oc get builds
NAME         TYPE     FROM          STATUS                       STARTED          DURATION
httpd-ex-1   Source   Git           Failed (FetchSourceFailed)   5 minutes ago    4s
httpd-ex-2   Source   Git@092fbaa   Complete                     40 seconds ago   36s

gabemontero:
> Launched cluster from this cluster, and trigger build from DPTP bc http://file.rdu.redhat.com/~hongkliu/dptp2021/bz1916897/content-mirror/app.ci.bc.content-mirror.yaml.
> 
> Build content-mirror-2 go to complete without /run/secret read only error.
> $ oc get builds -n ci
> NAME TYPE FROM STATUS STARTED DURATION
> content-mirror-2 Docker Git@083b29e Complete About an hour ago 1m16s
> 
> Build httpd-ex-2 with source secret go to complete too.
> $oc get builds
> NAME TYPE FROM STATUS STARTED DURATION
> httpd-ex-1 Source Git Failed (FetchSourceFailed) 5 minutes ago 4s
> httpd-ex-2 Source Git@092fbaa Complete 40 seconds ago 36s

thanks for the confirmation @xiuwang ... I had trouble either launching or updating clusters yesterday and was unable to do this

@nalind @bparees and now just need to get consensus on the correct way to apply this fix.

gabemontero:
sig-api-machinery flakes in latest e2e-aws

/test e2e-aws

pkg/build/builder/daemonless.go
gabemontero: > As you've guessed, persisting that across the build is not a feature that's built-in. I've just noticed that it persists this way with `ImageOptimizationSkipLayers`, but I think that's a bug.

So to try and take ^^ @nalind to the next step, should the plan of action be:
- fix the `ImageOptimizationSkipLayers` "bug" that persists the `/run/secrets/rhsm` (assuming we change the mount above to that)
- if so where is that code?  buildah? containers/image ? somewhere else ?
- then go with this change, as that will get us our desire that @bparees articulated above of only having writes affect the container's copy and not future containers or containers running concurrently

nalind: My understanding is that we want that to persist across a given stage in the Dockerfile, i.e., to make the way this accidentally works in `ImageOptimizationSkipLayers` cases be the behavior we provide all the time.

buildah's CLI provides an option for running stages in parallel, and trying to share a single mount across multiple stages could unexpectedly break builds that expected the data to persist across stages, but which implicitly depended on changes made in one stage being carried into subsequent stages.  My thinking, then, is that attempting to share a set of secrets across multiple stages in builds that use multiple stages is not desirable.

The location we use for storing the copy of the secrets is currently tied to the lifetime of the rootfs it's being used with, and the rootfs gets replaced each time buildah commits a layer.  In the `ImageOptimizationNone` case, that usually happens multiple times in a stage.  Storing the secrets in a location that gets reused across each step in a build stage would require a patch to the logic in buildah that determines where they go, since it currently doesn't expose any API options around controlling that.  I don't think we'd want to start exposing an option to control that in the API, since we'd want the behavior to be consistent between CLIs like podman and builds in OpenShift, and exposing an option at the API level would invite exposing it in CLIs, and that is not something I'd personally want to encourage.

gabemontero: OK @nalind  ... so I'm going to try and translate that into precise code changes in the PR:
- both commits currently associated with this PR are not good as is
- it sounds like we should still remove this code block, which the most recent commit https://github.com/openshift/builder/pull/206/commits/145e6513def3fe2655993573d2e48234a6707cae does, but rather than bringing back mounts.conf and mounting /run/screts/rhsm, we need to copy those contents into container only storage, and somehow point buildah to that container only storage .... that would also reconcile with @bparees 's https://github.com/openshift/builder/pull/206#discussion_r560497593  ... and would get back to one of the options I listed in slack yesterday

Correct ?

nalind: The "create a copy and bind mount to that instead" logic is part of how `mounts.conf` entries are currently handled.  The thing I'd be looking to change is how long a particular copy is kept and reused.  The current behavior, which varies in a way that's affected by the optimization policy setting, is more an accident than an intentional design decision.

gabemontero: > The "create a copy and bind mount to that instead" logic is part of how `mounts.conf` entries are currently handled. The thing I'd be looking to change is how long a particular copy is kept and reused. The current behavior, which varies in a way that's affected by the optimization policy setting, is more an accident than an intentional design decision.

So can changing "how long a particular copy is kept and reused" in an intentional way instead of an accidental way be accomplished with the current containers/* API, and we just need to change openshift/builder code, or do we need help from you and/or others on the podman/buildah/etc. team to get what is required ?

Sorry, even with your detailed explanations, I am unable to answer that question for myself.

nalind: The logic that controls where the copy of the secrets goes is currently hard-coded in buildah, so the patch would at least include a change there.  Changing it in a way that required any changes in openshift/builder would also require similar changes in podman and in any other consumers of the library to keep the behavior consistent between them, so my inclination would be to keep the decision making around that as an implementation detail in buildah.

bparees: I think i need to take back what i said earlier about it being ok if each RUN/layer gets a fresh copy of /run/secrets/rhsm, because the way people use that dir is probably:

COPY somesecret /run/secrets/rhsm
RUN yum install something-that-needs-the-above-secret

so either we fix buildah to persist the writes between layers (even when using using the layer optimization mode) or we add logic that copies the content to a new location before invoking buildah, and then bindmounts it.

gabemontero: > I think i need to take back what i said earlier about it being ok if each RUN/layer gets a fresh copy of /run/secrets/rhsm, because the way people use that dir is probably:
> 
> COPY somesecret /run/secrets/rhsm
> RUN yum install something-that-needs-the-above-secret
> 
> so either we fix buildah to persist the writes between layers (even when using using the layer optimization mode) 

so does @bparees 's comment ^^ lead you to modifying your comment at https://github.com/openshift/builder/pull/206#discussion_r561256117 @nalind ?

> or we add logic that copies the content to a new location before invoking buildah, and then bindmounts it.

So if possible @bparees I'd like to get a little more specific on how ^^ would be done to make sure we have a common understanding.

I'm going to attempt a mixture of real code and pseudo code to get a good detail vs. length mix.

```
startingDir := "/tmp"   // or some other location we agree upon
tmpDir, err := ioutil.TempDir(startingDir, "rhsm-creds")
if err != nil {
  ...
}

// logic to do recursive copy of /run/secrets/rhsm to tmpDir
...

// reintroduce transient mounts path 
...
transientMounts = append(transientMounts, fmt.Sprintf("%s:/run/secrets/rhsm:rw,nodev,noexec,nosuid", tmpDir))

```

how far off is that ?

bparees: that looks pretty close to what i had in mind.

gabemontero: OK @bparees @nalind @xiuwang I've dropped the re-introduction of `mounts.conf` and pushed a new commit that does the tmpdir/compy/transient mount @bparees and I just discussed 

Let's see how the e2e's go.

I'll try to run dptp's BC against a cluster-bot cluster with this PR (hopefully I have better luck with cluster-bot today), but if you get cycles @xiuwang by all means give it a go as well

@adambkaplan FYI  (to the whole PR :-) )

pkg/build/builder/daemonless.go
```diff
@@ -212,6 +212,25 @@ func daemonlessProcessLimits() (defaultProcessLimits []string) {
 	return defaultProcessLimits
 }
 
+func copyRHSMData(source, destination string) error {
+	var err error = filepath.Walk(source, func(path string, info os.FileInfo, err error) error {
+		var relPath string = strings.Replace(path, source, "", 1)
```
nalind, line 217: Should this be calling `filepath.Rel(source, path)`?

pkg/build/builder/daemonless.go
```diff
@@ -253,10 +272,21 @@ func buildDaemonlessImage(sc types.SystemContext, store storage.Store, isolation
...
+	if st, err := os.Stat("/run/secrets/rhsm"); err == nil && st.IsDir() {
+		// Add a bind of /run/secrets/rhsm, to pass along anything that the
+		// runtime mounted from the node into our /run/secrets/rhsm.
+		log.V(0).Infof("Adding transient rw bind mount for /run/secrets/rhsm")
+		tmpDir, err := ioutil.TempDir("/tmp", "rhsm-copy")
+		if err != nil {
+			log.V(0).Infof("Error creating tmpdir to set up /run/secrets/rhsm in build container: %s", err.Error())
+			return err
+		}
+		err = copyRHSMData("/run/secrets/rhms", tmpDir)
```
nalind, line 284: Typo: rhms -> rhsm

gabemontero: oops will fix thanks

pkg/build/builder/daemonless.go
```diff
@@ -212,6 +212,25 @@ func daemonlessProcessLimits() (defaultProcessLimits []string) {
 	return defaultProcessLimits
 }
 
+func copyRHSMData(source, destination string) error {
```
adambkaplan, line 215: We have similar "copy directory contents" in https://github.com/openshift/builder/blob/master/cmd/main.go#L74-L126. Can we consolidate into a utility package?

adambkaplan (changes requested):

pkg/build/builder/daemonless.go
gabemontero: > We have similar "copy directory contents" in https://github.com/openshift/builder/blob/master/cmd/main.go#L74-L126. Can we consolidate into a utility package?

update for this ^^ pushed @adambkaplan

gabemontero:
api server connection refused on image eco

/test e2e-aws-image-ecosystem

pkg/build/builder/daemonless.go
nalind: Do we test this with directories that contain symlinks to directories?  I think `ioutil.ReadDir()` uses `lstat()`, which would return `os.FileInfo`s that returns `false` from `IsDir()`.

gabemontero: Yeah this aspect came up when researching golang recursive copy, but I stopped short of adding that complexity since `/run/secret/rhsm` does not have symlinks based on my testing, so I stopped short of changing this existing code path.

But I'll defer to you @nalind and @adambkaplan on if we want to add this in just in case.

gabemontero:
OK tests are passing, and I was able to verifiy DPTP's BC using an image build from this commit

There is the one remaining discussion thread on the copy and symlinks, and then squashing the commits

gabemontero:
I've pushed a separate commit for symlinks @nalind @adambkaplan ... used https://stackoverflow.com/questions/51779243/copy-a-folder-in-go as a ref

gabemontero:
sig-network failures e2e-aws

/test e2e-aws

gabemontero:
sig-network flakes in e2e-aws

/retest

gabemontero:
/assign @adambkaplan 
/assign @nalind 

I'll defer to @bparees if he wants to now be unassigned

pkg/build/builder/util.go
```diff
@@ -150,3 +153,69 @@ func ParseProxyURL(proxy string) (*url.URL, error) {
 
 	return proxyURL, err
 }
+
+// CopyDirIfExists recursively copies a directory to the destination path.
+// If the source directory does not exist, no error is returned.
+// If the destination directory exists, any contents with matching file names
+// will be overwritten.
+func CopyDirIfExists(src, dst string) error {
```
adambkaplan, line 161: Please add unit tests for this function, covering in particular:

1. Source directory not existing
2. Destination already existing (confirm files overwritten
3. Regular files copied
4. Symlinks copied.

pkg/build/builder/util.go
```diff
@@ -150,3 +153,69 @@ func ParseProxyURL(proxy string) (*url.URL, error) {
...
+	_, err = io.Copy(out, in)
+	if err != nil {
+		return err
+	}
+	return out.Close()
+}
+
+// CopySymLink preserves symlinks as the form of copy
+func CopySymLink(src, dst string) error {
+	link, err := os.Readlink(src)
```
adambkaplan, line 216: What if the symlink refers to an absolute path, instead of a relative path? Do we keep as is?

See also https://github.com/openshift/source-to-image/blob/2ed02f2351b48858b5cc358bbccf6fc4a2a284ef/pkg/util/fs/fs.go#L208-L229

pkg/build/builder/util.go
```diff
@@ -150,3 +153,69 @@ func ParseProxyURL(proxy string) (*url.URL, error) {
...
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// CopyFileIfExists copies the source file to the given destination, if the source file exists.
+// If the destination file exists, it will be overwritten and will not copy file attributes.
+func CopyFileIfExists(src, dst string) error {
```
adambkaplan, line 190: Please add a unit test for this function.

adambkaplan (changes requested):
Something else to consider - we already have a "copy files & directories" library within source-to-image. Do we use that instead?

The default implementation copies symlinks by following the link and preserving the content.

https://github.com/openshift/source-to-image/tree/2ed02f2351b48858b5cc358bbccf6fc4a2a284ef/pkg/util/fs

gabemontero:
PR updated @adambkaplan to using s2i libs ... also went ahead and squashed commits - thanks!

gabemontero:
All green tests on our readonly /run PR @adambkaplan 

Reminder: follow up e2e's in https://github.com/openshift/origin/pull/25810 once this merges

adambkaplan:
/test e2e-aws-proxy

Want to make sure we didn't break the proxy CA generation, which I believe is utilized by git clone.

gabemontero:
e2e-aws-proxy succeeded @adambkaplan

adambkaplan (approved):
/lgtm

Once this merges we can verify with https://github.com/openshift/origin/pull/25810

PR #207, opened 2021-01-21
Force use of runc, since we know we're privileged, and customize our seccomp filter

gabemontero:
manual pick of @nalind 's https://github.com/openshift/builder/pull/202 required

/assign @bparees 
/assign @nalind 

@adambkaplan @jupierce FYI

PR #208, opened 2021-01-21
Force use of runc, since we know we're privileged, and customize our seccomp filter

gabemontero:
manual pick of @nalind 's https://github.com/openshift/builder/pull/207 required

/assign @adambkaplan 
/assign @nalind 
/assign @crawford 
/assign @jupierce 

@bparees FYI

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,7 @@ RUN hack/build.sh
 FROM registry.svc.ci.openshift.org/openshift/origin-v4.0:base
 # TODO: Add fuse-overlayfs once we build off of RHEL-8 UBI
 RUN INSTALL_PKGS=" \
-      bind-utils bsdtar findutils git hostname lsof socat \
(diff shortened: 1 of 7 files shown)


adambkaplan (commented):
@nalind will the seccomp profile work on a RHEL7-based container?

nalind:
> @nalind will the seccomp profile work on a RHEL7-based container?

I think we get warnings about syscalls mentioned in the policy if they aren't recognized by the version of the libseccomp C library we link with, but that's about it.  I don't think the kernel complains if a compiled filter references a syscall number that it doesn't implement.

gabemontero:
e2e-aws-build passed ... good news

image eco failure fix just merged recently

/retest

gabemontero:
more lease pain with e2e-aws

/retest

adambkaplan (approved):
/lgtm

PR #213, opened 2021-01-27
Bug 1905095: bump github.com/containers/image to v5.10.1

nalind:
Bump github.com/containers/image from v5.5.2, used in buildah 1.16.4, to v5.10.1, which extends the blob info cache to remember what type of compression was used for a given blob.  This also pulls in newer versions of some of its dependencies.


adambkaplan (approved):
/approve

Want to ensure that the test to verify fails before adding lgtm.

nalind:
/retitle Bug 1905095: bump github.com/containers/image to v5.10.0

nalind:
/retitle Bug 1905095: bump github.com/containers/image to v5.10.1

PR #218, opened 2021-02-10
Bug 1895053: Mount CA trust store in builds

adambkaplan:
Add the build container's CA trust store as a transient mount for buildah
if the `BUILD_MOUNT_ETC_PKI_CATRUST` environment variable is set to a
"true" value.


pkg/build/builder/daemonless.go
```diff
@@ -313,10 +301,77 @@ func buildDaemonlessImage(sc types.SystemContext, store storage.Store, isolation
...
+		return mounts, err
+	}
+	return mounts, nil
+}
+
+func appendRHSMMount(mounts []string) ([]string, error) {
+	st, err := os.Stat("/run/secrets/rhsm")
+	if err != nil {
+		return mounts, errors.Wrap(err, "failed to stat /run/secrets/rhsm")
+	}
```
adambkaplan, line 325: @gabemontero this is a refactor of your original code. However, I think we should not return an error.

From our experience, consuming entitlements in builds is more of a "secondary" use case. The worst case is if a build does a `yum install` and we don't provide the entitlement keys, the build fails at yum install time. I don't think we should fail all builds because we can't set up the rhsm data.

pkg/build/builder/daemonless.go
```diff
@@ -313,10 +301,77 @@ func buildDaemonlessImage(sc types.SystemContext, store storage.Store, isolation
...
+func appendCATrustMount(mounts []string) ([]string, error) {
+	mountCAEnv, exists := os.LookupEnv("BUILD_MOUNT_ETC_PKI_CATRUST")
+	if !exists {
+		return mounts, nil
+	}
+
+	mountCA, err := strconv.ParseBool(mountCAEnv)
+	if err != nil {
+		log.V(5).Infof("failed to parse BUILD_MOUNT_ETC_PKI_CATRUST: %v", err)
+		return mounts, errors.Wrap(err, "failed to parse BUILD_MOUNT_ETC_PKI_CATRUST")
```
adambkaplan, line 355: Likewise I don't think we should fail a build if we can't parse this env var or we don't find the trust bundle. If a build depends on this information, it will eventually fail.

adambkaplan (commented):
I'm think we shouldn't fail fast if we can't configure the transient mounts. Thoughts?

pkg/build/builder/daemonless.go
```diff
@@ -313,10 +301,77 @@ func buildDaemonlessImage(sc types.SystemContext, store storage.Store, isolation
...
+	if err != nil {
+		log.V(5).Infof("failed to stat /etc/pki/ca-trust: %v", err)
+		return mounts, errors.Wrap(err, "failed to stat /etc/pki/ca-trust")
+	}
+	if !st.IsDir() {
+		return mounts, nil
+	}
+
+	log.V(0).Infof("Adding transient ro bind mount for /etc/pki/ca-trust")
+	mounts = append(mounts, "/etc/pki/ca-trust:/etc/pki/ca-trust:ro")
```
adambkaplan, line 371: @nalind is this the right setting for a read-only mount? What do the `nodev,noexec,nosuid` arguments do for the rhsm mount?

nalind: Yes, that's the right flag for making the bind read-only.  The `nodev,noexec,nosuid` arguments request that in the new mountpoint, device nodes not be accessible, binaries not be executable despite the execute bit on their permissions, and that the setuid/setgid bit and file capabilities not be honored.

pkg/build/builder/daemonless.go
gabemontero: sure I'm good with that @adambkaplan

pkg/build/builder/daemonless.go
```diff
@@ -313,10 +301,77 @@ func buildDaemonlessImage(sc types.SystemContext, store storage.Store, isolation
...
+
+func appendCATrustMount(mounts []string) ([]string, error) {
+	mountCAEnv, exists := os.LookupEnv("BUILD_MOUNT_ETC_PKI_CATRUST")
+	if !exists {
+		return mounts, nil
+	}
+
+	mountCA, err := strconv.ParseBool(mountCAEnv)
+	if err != nil {
+		log.V(5).Infof("failed to parse BUILD_MOUNT_ETC_PKI_CATRUST: %v", err)
```
gabemontero, line 354: I'm fine with not failing on the parse error vs. later on @adambkaplan but if so I would prefer our clue/log here is not V(5) .... IMO a V(0) bread crumb is appropriate

pkg/build/builder/daemonless.go
```diff
@@ -313,10 +301,77 @@ func buildDaemonlessImage(sc types.SystemContext, store storage.Store, isolation
...
+		return mounts, errors.Wrap(err, "failed to parse BUILD_MOUNT_ETC_PKI_CATRUST")
+	}
+	if !mountCA {
+		return mounts, nil
+	}
+
+	st, err := os.Stat("/etc/pki/ca-trust")
+	if err != nil {
+		log.V(5).Infof("failed to stat /etc/pki/ca-trust: %v", err)
```
gabemontero, line 363: I am fine with this staying v(5) .... presumably it is not as susceptible to user error

adambkaplan:
@gabemontero this is ready for final review.

adambkaplan:
/retest

Builds hit known kubelet flake.

PR #220, opened 2021-03-09
unprivileged: add CLI options for isolation and storage

nalind:
Add CLI flags for controlling the type of process isolation we use for `RUN` instructions, and for controlling the storage driver and options that we pass to it, along with logic for guessing the right defaults.  Drop environment variables that controlled these settings.  This renders the static `storage.conf` file in the builder image largely irrelevant.
    
~~Add `/etc/subuid` and `/etc/subgid` files to the builder image that allow root to map the entire ID namespace.~~
    
Have the docker and sti builders re-exec themselves in user namespaces when they're run as non-UID=0 or without the `CAP_SYS_ADMIN` capabiity, so that they will be able to create new namespaces when processing RUN instructions.

With no CLI flags supplied and in a privileged pod, the builder should continue using the overlay driver using the kernel's overlay filesystem, and the OCI runtime.

### DIFF
--- a/cmd/builder.go
+++ b/cmd/builder.go
@@ -11,6 +11,7 @@ import (
 	kcmdutil "k8s.io/kubectl/pkg/cmd/util"
 	"k8s.io/kubectl/pkg/util/templates"
 
+	"github.com/containers/common/pkg/config"
 	"github.com/openshift/builder/pkg/build/builder/cmd"
 	"github.com/openshift/builder/pkg/version"
 )
@@ -65,15 +66,35 @@ func NewCmdVersion(fullName string, versionInfo k8sversion.Info, buildahVersion 
 
 // NewCommandS2IBuilder provides a CLI handler for S2I build type
 func NewCommandS2IBuilder(name string) *cobra.Command {
+	var isolation, ociRuntime, storageDriver, storageOptions string
+
+	defaultConfig, err := config.DefaultConfig()
+	kcmdutil.CheckErr(err)
+
 	cmd := &cobra.Command{
 		Use:   name,
 		Short: "Run a Source-to-Image build",
 		Long:  s2iBuilderLong,
 		Run: func(c *cobra.Command, args []string) {
-			err := cmd.RunS2IBuild(c.OutOrStderr())
+			var err error
+			if isolation == "" {
+				isolation, err = builderDefaultIsolation()
+				kcmdutil.CheckErr(err)
+			}
+			if storageDriver == "" {
+				storageDriver, storageOptions, err = builderDefaultStorage()
+				kcmdutil.CheckErr(err)
+			}
+			err = cmd.RunS2IBuild(c.OutOrStderr(), isolation, ociRuntime, storageDriver, storageOptions)
 			kcmdutil.CheckErr(err)
 		},
 	}
+
+	flags := cmd.Flags()
+	flags.StringVar(&isolation, "isolation", isolation, "type of process `isolation` to use for RUN instructions")
+	flags.StringVar(&ociRuntime, "oci-runtime", defaultConfig.Engine.OCIRuntime, "runtime to invoke for OCI isolation")
+	flags.StringVar(&storageDriver, "storage-driver", storageDriver, "storage driver to use for storing layers, images, and working containers")
+	flags.StringVar(&storageOptions, "storage-options", storageOptions, "storage options to use when storing layers, images, and working containers")
 
 	cmd.AddCommand(NewCmdVersion(name, version.Get(), version.BuildahVersion(), os.Stdout))
 	return cmd
@@ -81,15 +102,36 @@ func NewCommandS2IBuilder(name string) *cobra.Command {
 
 // NewCommandDockerBuilder provides a CLI handler for Docker build type
 func NewCommandDockerBuilder(name string) *cobra.Command {
+	var isolation, ociRuntime, storageDriver, storageOptions string
+
+	defaultConfig, err := config.DefaultConfig()
+	kcmdutil.CheckErr(err)
+
 	cmd := &cobra.Command{
 		Use:   name,
 		Short: "Run a Docker build",
 		Long:  dockerBuilderLong,
 		Run: func(c *cobra.Command, args []string) {
-			err := cmd.RunDockerBuild(c.OutOrStderr())
+			var err error
+			if isolation == "" {
+				isolation, err = builderDefaultIsolation()
+				kcmdutil.CheckErr(err)
+			}
+			if storageDriver == "" {
+				storageDriver, storageOptions, err = builderDefaultStorage()
+				kcmdutil.CheckErr(err)
+			}
+			err = cmd.RunDockerBuild(c.OutOrStderr(), isolation, ociRuntime, storageDriver, storageOptions)
 			kcmdutil.CheckErr(err)
 		},
 	}
+
+	flags := cmd.Flags()
+	flags.StringVar(&isolation, "isolation", isolation, "type of process `isolation` to use for RUN instructions")
+	flags.StringVar(&ociRuntime, "oci-runtime", defaultConfig.Engine.OCIRuntime, "runtime to invoke for OCI isolation")
+	flags.StringVar(&storageDriver, "storage-driver", storageDriver, "storage driver to use for storing layers, images, and working containers")
+	flags.StringVar(&storageOptions, "storage-options", storageOptions, "storage options to use when storing layers, images, and working containers")
+
 	cmd.AddCommand(NewCmdVersion(name, version.Get(), version.BuildahVersion(), os.Stdout))
 	return cmd
 }
--- a/cmd/defaults.go
+++ b/cmd/defaults.go
@@ -1,0 +1,106 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+
+	"github.com/containers/storage"
+	"k8s.io/klog/v2"
+)
+
+// Return "chroot" if we know we're not actually root, "oci" otherwise.
+func builderDefaultIsolation() (string, error) {
+	if inUserNamespace() {
+		// We probably don't have enough privileges to use a proper
+		// runtime.
+		// Lean on the container that we're in being itself
+		// unprivileged (i.e., having control groups including the
+		// device cgroup configured for us, being provided with a
+		// smaller set of devices in /dev, and likely running without a
+		// few capabilities that we don't need), and reduce the degree
+		// of isolation that we try to use to what we know we're
+		// actually allowed to do in an unprivileged container.
+		return "chroot", nil
+	}
+	// Use the proper runtime.
+	return "oci", nil
+}
+
+// Check that /dev/fuse is usable and we have the fuse-overlayfs helper.
+func builderCanUseOverlayFUSE() error {
+	device, err := os.Open("/dev/fuse")
+	if err != nil {
+		return fmt.Errorf("error opening device: %v", err)
+	}
+	defer device.Close()
+	if _, err := os.Stat("/usr/bin/fuse-overlayfs"); err != nil {
+		return err
+	}
+	return nil
+}
+
+// Try various storage setups until we find one that works with the privileges
+// that we currently have.
+func builderDefaultStorage() (string, string, error) {
+	for _, candidate := range []struct {
+		driver, options string
+		also            func() error
+	}{
+		{"overlay", `["mountopt=metacopy=on"]`, nil},
+		{"overlay", ``, nil},
+		{"overlay", `["mount_program=/usr/bin/fuse-overlayfs","mountopt=metacopy=on"]`, builderCanUseOverlayFUSE},
+		{"overlay", `["mount_program=/usr/bin/fuse-overlayfs"]`, builderCanUseOverlayFUSE},
+		{"vfs", "", nil},
+	} {
+		var options []string
+		// Is there an additional test?
+		if candidate.also != nil {
+			if why := candidate.also(); why != nil {
+				klog.V(2).Info(why.Error())
+				continue
+			}
+		}
+		// Are there options for this case?
+		if candidate.options != "" {
+			err := json.Unmarshal([]byte(candidate.options), &options)
+			if err != nil {
+				klog.Errorf("internal error parsing options %q: %v", candidate.options, err)
+				continue
+			}
+		}
+		// Precreate some things.
+		if _, err := os.Stat(fmt.Sprintf("/var/lib/shared/%s-layers/layers.lock", candidate.driver)); err == nil {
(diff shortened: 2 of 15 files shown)


gabemontero (commented):
@nalind @adambkaplan - perhaps I missed it in scrum, but is this change the final step for the builder image at least (i.e. ignore OCM changes we would make at some point, or global build config defaults) to support unprivileged builds?  weren't there some k8s changes we were waiting on as well?  if so, are those now there? or were those k8s changes for caching (I've forgotten these details more than once before :-) )

Should this PR be assoicated with https://issues.redhat.com/browse/BUILD-119

all in all, auspicious stuff :-)

pkg/build/builder/daemonless.go
```diff
@@ -615,15 +617,26 @@ type DaemonlessClient struct {
...
+func GetDaemonlessClient(systemContext types.SystemContext, store storage.Store, blobCacheDirectory string, isolationSpec string, imageOptimizationPolicy buildapiv1.ImageOptimizationPolicy) (client DockerClient, err error) {
 	if blobCacheDirectory != "" {
 		log.V(0).Infof("Caching blobs under %q.", blobCacheDirectory)
 	}
 
+	var isolation buildah.Isolation
+	switch isolationSpec {
+	case "", "oci":
+		isolation = buildah.IsolationOCI
+	case "chroot":
```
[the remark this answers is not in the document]
nalind: chrooting doesn't provide the isolation that you'd expect from a container, so a process we execute using it has privileges similar to the process that called it.  When called from inside of an unprivileged container, the unprivilegedness of that container provides the isolation we'd expect.

nalind:
When we're launched with privileged: false, we don't have enough privileges to run runc.

> @nalind @adambkaplan - perhaps I missed it in scrum, but is this change the final step for the builder image at least (i.e. ignore OCM changes we would make at some point, or global build config defaults) to support unprivileged builds? weren't there some k8s changes we were waiting on as well? if so, are those now there? or were those k8s changes for caching (I've forgotten these details more than once before :-) )

The build controller would still need to set "privileged: false" for the build pod, and optionally specify what type of isolation to use using the CLI "--isolation" flag to override the default that this PR adds.

> Should this PR be assoicated with https://issues.redhat.com/browse/BUILD-119

If "rootless" there means "uid != 0" from the node's point of view, then no, this doesn't get us that.

pkg/build/builder/daemonless.go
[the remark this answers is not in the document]
gabemontero: ok, so to map that back to my wording, chrooting when we are unprivileged avoids the pitfalls the CVE unveiled

agree with ^^ @nalind ?

gabemontero:
> When we're launched with privileged: false, we don't have enough privileges to run runc.
> 
> > @nalind @adambkaplan - perhaps I missed it in scrum, but is this change the final step for the builder image at least (i.e. ignore OCM changes we would make at some point, or global build config defaults) to support unprivileged builds? weren't there some k8s changes we were waiting on as well? if so, are those now there? or were those k8s changes for caching (I've forgotten these details more than once before :-) )
> 
> The build controller would still need to set "privileged: false" for the build pod, and optionally specify what type of isolation to use using the CLI "--isolation" flag to override the default that this PR adds.

Yeah that is the OCM related changes I'm referring to

> 
> > Should this PR be assoicated with https://issues.redhat.com/browse/BUILD-119
> 
> If "rootless" there means "uid != 0" from the node's point of view, then no, this doesn't get us that.

I honestly don't remember what was meant by "rootless" when Mrunal opened BUILD-119 and when @siamaksade opened https://issues.redhat.com/browse/BUILD-141

I'll defer to @adambkaplan

pkg/build/builder/daemonless.go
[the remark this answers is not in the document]
nalind: Not having a device control group set up is not an issue we have to avoid.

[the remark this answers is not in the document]
gabemontero: ok that sounds like the "why" for "chrooting when we are unprivileged avoids the pitfalls the CVE unveiled"

gabemontero:
e2e-aws-builds ->  `Cannot exceed quota for InstanceProfilesPerAccount: 1000`

cluster could not get off the ground

gabemontero:
@nalind you'll need https://github.com/openshift/origin/pull/25985 to merge in order to get a passing image-ecosystem

nalind:
Rebased.

nalind:
Rebased.  This may solve bug 1937069, where "chown" during a build takes a while, which I suspect due to copy-up that can be avoided if we use the metacopy feature, which this PR attempts to detect and use if it's supported.

cmd/defaults.go
```diff
@@ -0,0 +1,98 @@
...
+	if inUserNamespace() {
+		// We probably don't have enough privileges to use a proper runtime.
+		return "chroot", nil
+	}
+	// Use the proper runtime.
+	return "oci", nil
+}
+
+// Check that /dev/fuse is usable and we have the fuse-overlayfs helper.
+func builderCanUseOverlayFUSE() string {
```
rhatdan, line 23: Why isn't this returning an error?

rhatdan:
One question, but overall LGTM

cmd/defaults.go
nalind: Sure, why not.  Changing it to return an error.

rhatdan:
/approve
LGTM

gabemontero:
per discussion in team scrum including with @nalind 

/assign @adambkaplan 

to minimally get the approve label on this, and if he has cycles decide whether it lgtm to him

I'll be taking another pass after posting this comment

cmd/defaults.go
```diff
@@ -0,0 +1,98 @@
...
+func builderDefaultStorage() (string, string, error) {
+	for _, candidate := range []struct {
+		driver, options string
+		also            func() error
+	}{
+		{"overlay", `["mountopt=metacopy=on"]`, nil},
+		{"overlay", ``, nil},
+		{"overlay", `["mount_program=/usr/bin/fuse-overlayfs","mountopt=metacopy=on"]`, builderCanUseOverlayFUSE},
+		{"overlay", `["mount_program=/usr/bin/fuse-overlayfs"]`, builderCanUseOverlayFUSE},
+		{"vfs", "", nil},
```
gabemontero, line 46: ah the order of precedence @nalind mentioned in office hours today :-)

cmd/main.go
```diff
@@ -44,27 +46,47 @@ func main() {
...
-		fmt.Printf("Error setting up cluster CA cert: %v", err)
-		os.Exit(1)
-	}
-
-	runtimeCASrc := fmt.Sprintf("%s/certs.d", builder.ConfigMapCertsMountPath)
-	err = fs.CopyContents(runtimeCASrc, runtimeCertRoot, map[string]string{})
-	if err != nil {
-		fmt.Printf("Error setting up service CA cert: %v", err)
-		os.Exit(1)
+	if !inUserNamespace() {
```
gabemontero, line 49: So why do we only do this when not in user namespaces?

How is this setup provided when we are in user namespaces?

Looking for a comment that answers those questions.  Thanks.

gabemontero (changes requested):
Aside from the request for a comment below @nalind ideally some added unit tests in https://github.com/openshift/builder/blob/master/pkg/build/builder/cmd/builder_test.go and https://github.com/openshift/builder/blob/master/pkg/build/builder/daemonless_test.go to validate that the new parameters being passed in are processed correctly seems warranted.

Otherwise looks great / thanks !!

cmd/main.go
nalind: If we've set up a user namespace, then we know that we weren't started as UID 0 in the container, so these locations are not writable.  If we try to write to those locations anyway, we hit a permissions error and exit.  It's an unsolved problem for cases where we start the container as a UID other than 0.

gabemontero: OK caputure that in a comment in the code and I'm good on this one

cmd/builder.go
```diff
@@ -65,31 +65,66 @@ func NewCmdVersion(fullName string, versionInfo k8sversion.Info, buildahVersion
...
+				kcmdutil.CheckErr(err)
+			}
+			err = cmd.RunS2IBuild(c.OutOrStderr(), isolation, ociRuntime, storageDriver, storageOptions)
 			kcmdutil.CheckErr(err)
 		},
 	}
 
+	flags := cmd.Flags()
+	flags.StringVar(&isolation, "isolation", isolation, "type of process `isolation` to use for RUN instructions")
+	flags.StringVar(&ociRuntime, "oci-runtime", ociRuntime, "runtime to invoke for OCI isolation")
```
adambkaplan, line 91: What happens if `--oci-runtime` is not specified? Is there a reasonable default for the empty string?

cmd/builder.go
```diff
@@ -65,31 +65,66 @@ func NewCmdVersion(fullName string, versionInfo k8sversion.Info, buildahVersion
...
+				kcmdutil.CheckErr(err)
+			}
+			err = cmd.RunDockerBuild(c.OutOrStderr(), isolation, ociRuntime, storageDriver, storageOptions)
 			kcmdutil.CheckErr(err)
 		},
 	}
+
+	flags := cmd.Flags()
+	flags.StringVar(&isolation, "isolation", isolation, "type of process `isolation` to use for RUN instructions")
+	flags.StringVar(&ociRuntime, "oci-runtime", ociRuntime, "runtime to invoke for OCI isolation")
```
adambkaplan, line 124: Same as above

cmd/defaults.go
```diff
@@ -0,0 +1,98 @@
...
+
+	"github.com/containers/storage"
+	"k8s.io/klog/v2"
+)
+
+// Return "chroot" if we know we're not actually root, "oci" otherwise.
+func builderDefaultIsolation() (string, error) {
+	if inUserNamespace() {
+		// We probably don't have enough privileges to use a proper runtime.
+		return "chroot", nil
```
adambkaplan, line 16: If we are in a user namespace, does this make us vulnerable to https://access.redhat.com/security/cve/CVE-2021-20182?

cmd/main.go
adambkaplan: So as non-root we can't write to `/etc/docker/certs.d`? Interesting, may need to determine how we can have the build controller mount these for us.

adambkaplan: Will this also impact our entrypoint script that generates the trust bundle?

adambkaplan (commented):
/approve

There are definite issues if we run in a user namespace around certs that need to be addressed. The scope is way beyond this PR and should be captured in JIRA.

cmd/defaults.go
nalind: In an unprivileged container, the builder container will already be have a device control group configured for limiting which devices the kernel will allow direct access to, even when the nodes exist, and `/dev` will have the smaller set of entries in it which we grant to unprivileged containers.  I expect we'll also be running without CAP_MKNOD.

cmd/main.go
nalind: Mounting the `certs.d` subpath of the `build-ca-bundles` volume to `/etc/docker/certs.d` _should_ work, since the directory isn't even in the builder image to begin with.  The `cluster.crt` that we drop in `/etc/pki/tls/certs` is trickier, since the directory already includes content that we want to still have there.  We should be able to mount the file itself from the volume that contains the service account secrets, but I haven't personally tried that.
This may all be moot, though, since I think ensuring the directories exist and setting them group-writeable in the image will also work, with no code changes.

cmd/defaults.go
gabemontero: fwiw @adambkaplan we also dove into the CVE ramifications back in March with https://github.com/openshift/builder/pull/220#discussion_r590595420

bottom line:  "we appear good"

that said, I think @nalind 's comment here ^^ would make a good comment when we see this code again in 6 months and ask the same question, not remembering this discussion :-)

gabemontero:
> /approve
> 
> There are definite issues if we run in a user namespace around certs that need to be addressed. The scope is way beyond this PR and should be captured in JIRA.

I'm good with that ^^ but for me this highlights the need for comments I noted earlier.  Even a TODO comment with a ref to whatever Jira gets opened for this.

cmd/builder.go
nalind: We can retrieve a default from `github.com/containers/common/pkg/config`'s `DefaultConfig()` function.  I'll add that.

cmd/defaults.go
nalind: Adding `Lean on the container that we're in being itself unprivileged (i.e., having control groups including the
device cgroup configured for us, being provided with a smaller set of devices in /dev, and likely running without a
few capabilities that we don't need), and reduce the degree of isolation that we try to use to what we know we're
actually allowed to do in an unprivileged container.`

cmd/main.go
nalind: Going to try creating the directories and setting them group-writable in the `Dockerfile` instead.  That'll probably work better for the entrypoint script, too.

cmd/defaults.go
gabemontero: thanks @nalind

cmd/main.go
gabemontero: OK if !inUserNamespace() condition has been removed along with the Dockerfile changes @nalind mentions ^^

cmd/builder.go
gabemontero: ok see that with https://github.com/openshift/builder/pull/220/files#diff-be76072c86dc24313ebc27eca7d613a1068dc337b068610f3f0e98f6e97ac5b0R71

cmd/builder.go
gabemontero: addressed via https://github.com/openshift/builder/pull/220/files#diff-be76072c86dc24313ebc27eca7d613a1068dc337b068610f3f0e98f6e97ac5b0R71 per @nalind

gabemontero:
last e2e-aws-builds run died in aws install 

all the prior comments have been addressed, so 

/lgtm

but let's keep an eye on whether possible future e2e-aws-build fail in the actual tests

also, given the fact that we are touching the cert path, I'm launching a 

/test e2e-aws-proxy

(pretty sure we made that an optional possibility with this repo)

gabemontero:
any hint what changed from f0950af to f1de8bd @nalind ?

gabemontero:
the one e2e-aws failure is suppose to be a low rate flake according to sippy with

Associated Bugs: 1972490 1973266 1978829

/test e2e-aws

nalind:
> any hint what changed from [f0950af](https://github.com/openshift/builder/commit/f0950afdb4440c8fc611bcc816bfc0eb72e7ed56) to [f1de8bd](https://github.com/openshift/builder/commit/f1de8bdfc9d702db91b515ae05bf77191bc3f70c) @nalind ?

That should have dropped a helper function that was introduced in an earlier version of this PR but which is no longer used.

PR #223, opened 2021-03-15
Bug 1939218: bump(containers/*):

adambkaplan:
Bump containers/image and containers/storage to fix blobinfocache bugs
that caused builds to push image manifests that did not conform to docker
v2schema2.


adambkaplan:
/close

Base commit was not correct.

PR #225, opened 2021-03-15
Bug 1939219: bump(*):

adambkaplan:
Bump containers/image and containers/storage to fix blobinfocache bugs
that caused builds to push image manifests that did not conform to docker
v2schema2. Removed the "replace" clause for containers/image because
buildah can support containers/image v5.10.5. Re-generated go.sum file
due to s2i changing its tag SHA for v1.3.0.


adambkaplan:
/hold

Needs openshift/origin#25967 to merge for builds suite to pass

nalind:
LGTM

wewang58:
e2e-aws-builds  job failed because using ruby 2.5 image to run ruby 2.7 command in dockerfile, so failed when [pr25967 ](https://github.com/openshift/origin/pull/25967) merge, it will pass  i think.

PR #236, opened 2021-03-29
Bug 1943614: add explict exit log after buildah pull image; tweak existing start log priot to buildah pull image call

gabemontero:
Our CI / testplatform team are seeing inexplicably long delays when OpenShift Builds try to pull image content as input prior to building the image.

Some corresponding "end" messages to go with the "start" messages that already exist was deemed helpful in their diagnostic efforts. so they easily demarcate when the openshift/builder image calls out into the vendored buildah/containers code.

/assign @bparees

### DIFF
--- a/pkg/build/builder/daemonless.go
+++ b/pkg/build/builder/daemonless.go
@@ -160,7 +160,7 @@ func mergeNodeCredentialsDockerAuth(credsPath string) *docker.AuthConfigurations
 }
 
 func pullDaemonlessImage(sc types.SystemContext, store storage.Store, imageName string, searchPaths []string, blobCacheDirectory string) error {
-	log.V(0).Infof("Asked to pull fresh copy of %q.", imageName)
+	log.V(2).Infof("Attempting pull of image %q.", imageName)
 
 	if imageName == "" {
 		return fmt.Errorf("unable to pull using empty image name")
@@ -201,9 +201,13 @@ func pullDaemonlessImage(sc types.SystemContext, store storage.Store, imageName 
 		BlobDirectory: blobCacheDirectory,
 	}
 	_, err = buildah.Pull(context.TODO(), "docker://"+imageName, options)
-	log.V(0).Infof("Finished pulling image %s", imageName)
+	if err == nil {
+		log.V(2).Infof("Finished pulling image %q", imageName)
+	} else {
+		log.V(2).Infof("Error pulling image %q: %s", imageName, err.Error())
+	}
 	if err != nil && dockerConfigCredsErr != nil {
-		err = fmt.Errorf("%s; also, error processing dockerconfigjson: %s", err.Error(), dockerConfigCredsErr.Error())
+		err = fmt.Errorf("Error pulling image %q: %s; also, error processing dockerconfigjson: %s", imageName, err.Error(), dockerConfigCredsErr.Error())
 	}
 	return err
 }


gabemontero:
`Post "http://ghproxy/graphql": dial tcp 172.30.229.2:80: i/o timeout ` on the bugzilla bot's attempt

will retry in a bit

pkg/build/builder/daemonless.go
```diff
@@ -160,7 +160,7 @@ func mergeNodeCredentialsDockerAuth(credsPath string) *docker.AuthConfigurations
 }
 
 func pullDaemonlessImage(sc types.SystemContext, store storage.Store, imageName string, searchPaths []string, blobCacheDirectory string) error {
-	log.V(0).Infof("Asked to pull fresh copy of %q.", imageName)
+	log.V(2).Infof("Asked to pull a fresh copy of %q.", imageName)
```
bparees, line 163:
```suggestion
	log.V(2).Infof("Attempting pull of image %q.", imageName)
```

bparees: I see this is back to v2.  Did you confirm we're printing v2 by default?  Or that the dptp team is prepared to configure the loglevel for their builds explicitly to get this output if they want it?

pkg/build/builder/daemonless.go
```diff
@@ -201,7 +201,7 @@ func pullDaemonlessImage(sc types.SystemContext, store storage.Store, imageName
 		BlobDirectory: blobCacheDirectory,
 	}
 	_, err = buildah.Pull(context.TODO(), "docker://"+imageName, options)
-	log.V(0).Infof("Finished pulling image %s", imageName)
+	log.V(2).Infof("Finished the attempt to pull a fresh copy of %q", imageName)
```
bparees, line 204:
```suggestion
        if err==nil {
   	   log.V(2).Infof("Finished pulling image %q", imageName)
	}
```

pkg/build/builder/daemonless.go
```diff
@@ -201,7 +201,7 @@ func pullDaemonlessImage(sc types.SystemContext, store storage.Store, imageName
...
 	if err != nil && dockerConfigCredsErr != nil {
 		err = fmt.Errorf("%s; also, error processing dockerconfigjson: %s", err.Error(), dockerConfigCredsErr.Error())
```
bparees, line 206:
```suggestion
		err = fmt.Errorf("Error pulling image %q: %s; also, error processing dockerconfigjson: %s", imageName,err.Error(), dockerConfigCredsErr.Error())
```

pkg/build/builder/daemonless.go
gabemontero: the latter, at least initially ... in fact, they have to go to V6 to get the info down in the buildah layer that they need

bparees: in that case while i think this is still worth cleaning up, it seems like there's no need to backport it for them?

gabemontero: actually the thought was having a start/end around calling into buildah would help both them, as well as build api team, to more easily delineate "we've dropped down into vendored containers code, we've come back up from vendored containers code"

I'd like to backport this to 4.7, as that is where testplatform will be investigating the current slowness, an soliciting build api and buildah teams for diagnostics

bparees: but it's not actually adding any net new logging they don't already get today (just changing the wording), so what's the value in backporting it?

they can already see the start/end of the pull w/ the existing logging.

gabemontero: that was not clear from the last time we looked at this, at least from the sense that there was not a mixing of build image concerns and buildah concerns in those start/end logs

there was a clear start log, but no end log that I saw, or see now looking at the code

but maybe I am missing it ... where do thing the explicit "end" log is for pulling ImageSource entries?

bparees: i mean this PR is not adding any net net log statements.  So either:

1) the existing log statements map to begin/end of the pull (even if the wording isn't perfect) and there's no need to backport the wording cleanup to 4.7

or

2) this PR isn't sufficient to address the ask (to add explicit start/end log statements)

what am i missing?

gabemontero: Well, upon review, I think it is fair to say the PR title is misleading.  The "start" log already exists.  Though we tweaked the wording with this PR.  So https://github.com/openshift/builder/pull/236/files#diff-9a18a571061b068db5c93ccfb3b354c59d32cb835224bbee929350034fb7fb92R163

now, in the original PR, I originally had a new log right before https://github.com/openshift/builder/pull/236/files#diff-9a18a571061b068db5c93ccfb3b354c59d32cb835224bbee929350034fb7fb92R203 but Adam said that was redundant to https://github.com/openshift/builder/pull/236/files#diff-9a18a571061b068db5c93ccfb3b354c59d32cb835224bbee929350034fb7fb92R163 in that nothing of consequence happened between those 2 lines

And this PR then adds the end log at https://github.com/openshift/builder/pull/236/files#diff-9a18a571061b068db5c93ccfb3b354c59d32cb835224bbee929350034fb7fb92R205

I don't see any other existing logs that demarcate ^^ explicitly ... that more than anything is what would help, so we can distinguish clearly when pulling the ImageSources end and the rest of the build commences.

I'll update the commit and PR title around "adding" exit, "tweaking" start messages.

Does that help?

bparees: > And this PR then adds the end log at https://github.com/openshift/builder/pull/236/files#diff-9a18a571061b068db5c93ccfb3b354c59d32cb835224bbee929350034fb7fb92R205

this PR isn't adding that line, though.  it's already there(this PR just adds a conditional around it and changes the loglevel).  Or you're saying that line isn't in 4.7 today?  that might be the detail i was missing.  If it's not in 4.7 today then yeah, backporting makes sense.

pkg/build/builder/daemonless.go
```diff
@@ -201,9 +201,11 @@ func pullDaemonlessImage(sc types.SystemContext, store storage.Store, imageName
 		BlobDirectory: blobCacheDirectory,
 	}
 	_, err = buildah.Pull(context.TODO(), "docker://"+imageName, options)
-	log.V(0).Infof("Finished pulling image %s", imageName)
+	if err == nil {
+		log.V(2).Infof("Finished pulling image %q", imageName)
+	}
 	if err != nil && dockerConfigCredsErr != nil {
-		err = fmt.Errorf("%s; also, error processing dockerconfigjson: %s", err.Error(), dockerConfigCredsErr.Error())
+		err = fmt.Errorf("Error pulling image %q: %s; also, error processing dockerconfigjson: %s", imageName, err.Error(), dockerConfigCredsErr.Error())
```
bparees, line 208: sorry i missed this earlier..... we only return this error if there is both a configjson error *and* an error pulling?  that seems weird.  (Though i realize it's not changed by this PR)

Also since i had you add the conditional to only print the finished statement when err==nil, i guess we also need to log something explicitly to demarcate completion when err!=nil.  (Rather than just relying on the caller to print out the error that we are returning here)

pkg/build/builder/daemonless.go
```diff
@@ -201,9 +201,11 @@ func pullDaemonlessImage(sc types.SystemContext, store storage.Store, imageName
 		BlobDirectory: blobCacheDirectory,
 	}
 	_, err = buildah.Pull(context.TODO(), "docker://"+imageName, options)
-	log.V(0).Infof("Finished pulling image %s", imageName)
+	if err == nil {
+		log.V(2).Infof("Finished pulling image %q", imageName)
+	}
```
bparees, line 206:
```suggestion
	} else {
	  log.Errorf("Error pulling image %q: %s",imageName,err.Error())
	}
```

pkg/build/builder/daemonless.go
gabemontero: yes sorry that is it ... we are "cleaning" up the logs we originally introduced in 4.8, per you comments on "making them professional" in the 4.7 PR, v2 is the default, etc.

once this merges, I'll manually pick this commit and the prior commit to make sure the final result here in 4.8 lands in 4.7

pkg/build/builder/daemonless.go
gabemontero: yeah weird but there is some reasoning behind it as I recall ... unexpected errors because of bad pull secrets vs. an error which we might be able to ignore if we retry with the node credentials

after I pull in your suggested error i'll add a conditional to cover this gap

pkg/build/builder/daemonless.go
gabemontero: oh wait you suggested change is for that gap

gabemontero: there is not `log.Errorf` ... just going to keep V(2) as we may be able to ingore this / recover on a retry

PR #240, opened 2021-04-01
Bug 1945692: move entitlement related secrets back to mounts.conf

gabemontero:
manual pick of https://github.com/openshift/builder/pull/238 needed

/assign @bparees

PR #239, opened 2021-03-31
Bug 1951084: remove mounts.conf again, but patch buildah change needed to make transient mounts work for us

gabemontero:
@nalind @bparees the "meets min" change to buildah that allows host entitlements to work for openshift builds, but without a mounts.conf file in the builder image, and warning message when we run on rhcos

https://github.com/containers/buildah/pull/3117 is a slightly different take

I'll next see about porting that in, and verify on our UPI with entitlements cluster

PR #241, opened 2021-04-04
remove mounts.conf again, but patch buildah change needed to make transient mounts work for us

gabemontero:
The 4.7 version of https://github.com/openshift/builder/pull/239
that vendor's in buildah's https://github.com/containers/buildah/pull/3117/commits that allows us to use transient mounts vs. mounts.conf so we don't get those benign warning messages from buildah when the entitlement file are unavailable on RHCOS

this would ultimately replace https://github.com/openshift/builder/pull/240 in 4.7

an options to consider, if we trust the manual verification of the buildah fix I did last week, or stop the merge of https://github.com/openshift/builder/pull/240 in 4.7 and just go with this.

@bparees @nalind @adambkaplan


gabemontero:
e2e-aws failure in ` kubelet terminates kube-apiserver gracefully `

/test e2e-aws

bparees:
i certainly like this fix better if we're confident in it.

adambkaplan:
@gabemontero do we have a pressing need to backport #239 to 4.7? If we've fixed the root issue already, then I'm not inclined to backport something that isn't user-facing.

gabemontero:
> @gabemontero do we have a pressing need to backport #239 to 4.7? If we've fixed the root issue already, then I'm not inclined to backport something that isn't user-facing.

One could claim it is user facing @adambkaplan because if the user looks at a build log and sees those Warning messages, they may get concerned.  And unless they think of looking at the release notes to confirm those warning messages are benign, perhaps they instead file a customer to case to ask or complain about them.

@bparees also wanted us to move toward this "cleaner" form long term.

WDYT?

If you are swayed, I'll open BZs for the 4.8 branch and this one, and "officially" start things.

thanks

gabemontero:
per discussion with @adambkaplan we'll do https://github.com/openshift/builder/pull/239 in master/4.8 but as a rule we do not bump buildah like this in the z stream so close this one out ... we'll go with the openshift/builder only form that has the benign warning messages when running on rhcos

PR #244, opened 2021-05-05
Bug 1925412: update dependencies affected by CVE-2021-3121

alicerum:
Some dependencies had to regenerate code by newer version of gogo/protobuf, and because of that we have to depend on newer version of them.

This dependencies have been updated:

* vendor/k8s.io/api
* vendor/k8s.io/apimachinery
* vendor/github.com/containerd/cgroups
* vendor/github.com/openshift/api

**Not going to update this:**
* vendor/github.com/docker/docker

`docker/docker` is replaced in our case by `openshift/moby-moby`. The only package we import which has code generated by `protobuf` is [swarm/runtime](https://github.com/openshift/moby-moby/tree/openshift-3.10-docker-b68221c/api/types/swarm/runtime).
We do not use this code in `openshift/builder`, and it is imported by references with other code that we use which is not generated by `protobuf`.
All the rest parts of code can be considered safe because they are generated by `protobuf` `v1.3.2`.


go.mod
```diff
@@ -7,59 +7,59 @@ require (
...
-	k8s.io/component-base v0.20.0
-	k8s.io/klog/v2 v2.4.0
-	k8s.io/kubectl v0.20.0
-	k8s.io/kubernetes v1.20.0
+	k8s.io/api v0.21.0-rc.0
+	k8s.io/apimachinery v0.21.0-rc.0
+	k8s.io/client-go v0.21.0-rc.0
+	k8s.io/component-base v0.21.0-rc.0
+	k8s.io/klog/v2 v2.8.0
+	k8s.io/kubectl v0.20.6
```
gabemontero, line 31: I would expect this to be bumped to v0.21.0-rc.0 as well

go.mod
```diff
@@ -7,59 +7,59 @@ require (
...
+	k8s.io/kubernetes v1.19.10
```
gabemontero, line 32: I would expect this to be bumped to v0.21.0-rc.0 as well

go.mod
```diff
@@ -7,59 +7,59 @@ require (
...
-	k8s.io/client-go v0.20.0
-	k8s.io/component-base v0.20.0
-	k8s.io/klog/v2 v2.4.0
-	k8s.io/kubectl v0.20.0
-	k8s.io/kubernetes v1.20.0
+	k8s.io/api v0.21.0-rc.0
+	k8s.io/apimachinery v0.21.0-rc.0
+	k8s.io/client-go v0.21.0-rc.0
+	k8s.io/component-base v0.21.0-rc.0
+	k8s.io/klog/v2 v2.8.0
```
gabemontero, line 30: of course klog is separate enough that this is fine

go.mod
```diff
@@ -7,59 +7,59 @@ require (
...
-	k8s.io/kube-controller-manager => k8s.io/kube-controller-manager v0.20.0
-	k8s.io/kube-proxy => k8s.io/kube-proxy v0.20.0
-	k8s.io/kube-scheduler => k8s.io/kube-scheduler v0.20.0
-	k8s.io/kubectl => k8s.io/kubectl v0.20.0
-	k8s.io/kubelet => k8s.io/kubelet v0.20.0
-	k8s.io/kubernetes => k8s.io/kubernetes v1.19.3
-	k8s.io/legacy-cloud-providers => k8s.io/legacy-cloud-providers v0.20.0
-	k8s.io/metrics => k8s.io/metrics v0.20.0
-	k8s.io/sample-apiserver => k8s.io/sample-apiserver v0.20.0
+	k8s.io/api => k8s.io/api v0.20.6
```
gabemontero, line 43: as a result of keeping k8s.io/kubernetes consistent with your bumps of other k8s deps to v0.21.0-rc.0, I would expect all of these to be hard coded to v0.21.0-rc.0

gabemontero (changes requested):
I think some devil in the details, probably undocumented, "best practices" around how we bump k8s apply here

go.mod
```diff
@@ -7,59 +7,59 @@ require (
...
 	github.com/spf13/pflag v1.0.5
 	golang.org/x/sys v0.0.0-20210324051608-47abb6519492
-	k8s.io/api v0.20.0
-	k8s.io/apimachinery v0.20.0
-	k8s.io/client-go v0.20.0
-	k8s.io/component-base v0.20.0
-	k8s.io/klog/v2 v2.4.0
-	k8s.io/kubectl v0.20.0
-	k8s.io/kubernetes v1.20.0
+	k8s.io/api v0.21.0-rc.0
```
adambkaplan, line 26: The "current" tag for k8s.io repos is `v0.21.0` - `-rc.0` was the release candidate.

go.mod
adambkaplan: This probably got auto-bumped with the k8s update.

go.mod
adambkaplan: For go.mod reasons k8s.io/kubernetes keeps the version numbers aligned with what we would expect for semvers. So the current version of k8s.io/kubernetes is v1.21.0

go.mod
```diff
@@ -7,59 +7,59 @@ require (
...
+	k8s.io/component-base => k8s.io/component-base v0.20.6
+	k8s.io/cri-api => k8s.io/cri-api v0.20.6
+	k8s.io/csi-translation-lib => k8s.io/csi-translation-lib v0.20.6
+	k8s.io/kube-aggregator => k8s.io/kube-aggregator v0.20.6
+	k8s.io/kube-controller-manager => k8s.io/kube-controller-manager v0.20.6
+	k8s.io/kube-proxy => k8s.io/kube-proxy v0.20.6
+	k8s.io/kube-scheduler => k8s.io/kube-scheduler v0.20.6
+	k8s.io/kubectl => k8s.io/kubectl v0.20.6
+	k8s.io/kubelet => k8s.io/kubelet v0.20.6
+	k8s.io/kubernetes => k8s.io/kubernetes v1.19.10
```
adambkaplan, line 61: I don't think we need a replace here (I _think_). Bumping k8s.io/kubernetes to v1.21.0 above should be sufficient.

adambkaplan (changes requested):
Once we're happy with the bump we should merge this PR. Patching our docker/docker fork should be done separately and merged in a different pull request so that we land a rebase.

go.mod
alicerum: Actually I wanted to keep it at `0.20.6`, but something keeps automatically bumping it up to `1.21` on `go mod vendor`. I'll investigate.

go.mod
alicerum: yeah, since newer `k8s` brings new `klog`, it got bumped here, too. if we need to keep it down, can add `replace`, but i don't think it's really necessary.

xiuwang:
Do builds regression test on the cluster launched from pr, no issue found.

/label qe-approved

adambkaplan (approved):
/lgtm

PR #245, opened 2021-05-11
BUILD-87: secret configmap volume mounts in builds

coreydaley:
https://issues.redhat.com/browse/BUILD-87


pkg/build/builder/daemonless.go
```diff
@@ -311,118 +328,178 @@ func buildDaemonlessImage(sc types.SystemContext, store storage.Store, isolation
...
 
-func generateTransientMounts() []string {
-	mounts := []string{}
-	mounts = appendRHSMMount(defaultMountStart, mounts)
-	mounts = appendETCPKIMount(defaultMountStart, mounts)
-	mounts = appendRHRepoMount(defaultMountStart, mounts)
-	mounts = appendCATrustMount(mounts)
+type TransientMounts map[string]string
+
+func (t TransientMounts) append(mount string) error {
```
adambkaplan, line 337: Why not accept an object that mirrors or extends the `BuildVolumeMount` object in the api? That way `append` can encode the data into the internal map, with the appropriate options added on.

```suggestion
type TransientMount struct {
   buildv1.BuildVolumeMount
   // additional buildah mount options not exposed to the Build API
  NoDev bool 
  NoExec bool
  NoSuid bool
}

func (t TransientMounts) append(mount TransientMount) error {
```

pkg/build/builder/daemonless.go
```diff
@@ -311,118 +328,178 @@ func buildDaemonlessImage(sc types.SystemContext, store storage.Store, isolation
...
+		var mountOptions []string
+		if bvm.ReadOnly {
+			mountOptions = append(mountOptions, "ro")
+		} else {
+			mountOptions = append(mountOptions, "rw")
+		}
+
+		if err := mountsMap.append(fmt.Sprintf("%s:%s:%s", filepath.Join(BuildVolumeMountPath, bvm.MountPath), bvm.MountPath, strings.Join(mountOptions, ","))); err != nil {
+			return err
+		}
```
adambkaplan, line 401: If `append` accepts a fuller TransientMount object as an argument, then this logic is not repeated.

pkg/build/builder/daemonless.go
```diff
@@ -311,118 +328,178 @@ func buildDaemonlessImage(sc types.SystemContext, store storage.Store, isolation
...
 	if !st.IsDir() {
 		log.V(0).Infof("custom PKI trust bundle will not be available in this build: /etc/pki/ca-trust is not a directory")
-		return mounts
+		return nil
 	}
 
 	log.V(0).Infof("Adding transient ro bind mount for /etc/pki/ca-trust")
-	mounts = append(mounts, "/etc/pki/ca-trust:/etc/pki/ca-trust:ro")
-	return mounts
+	return mountsMap.append("/etc/pki/ca-trust:/etc/pki/ca-trust:ro")
```
adambkaplan, line 502: Ditto - TransientMount object will abstract away the read-only bit.

adambkaplan (changes requested):

pkg/build/builder/daemonless.go
```diff
@@ -311,118 +328,233 @@ func buildDaemonlessImage(sc types.SystemContext, store storage.Store, isolation
...
+	NoDev  bool
+	NoExec bool
+	NoSuid bool
+}
+
+func (tm TransientMount) String() string {
+	var mountOptions []string
+	if !tm.BuildVolumeMount.ReadOnly {
+		mountOptions = append(mountOptions, "rw")
+	}
```
adambkaplan, line 348: Need an else for read-only
```suggestion
	if !tm.BuildVolumeMount.ReadOnly {
		mountOptions = append(mountOptions, "rw")
	} else {
		mountOptions = append(mountOptions, "ro") 
	}
```

pkg/build/builder/daemonless.go
```diff
@@ -311,118 +328,233 @@ func buildDaemonlessImage(sc types.SystemContext, store storage.Store, isolation
...
+		buildVolumeMounts = build.Spec.Strategy.DockerStrategy.VolumeMounts
+	}
+
+	for _, bvm := range buildVolumeMounts {
+		var mountOptions []string
+		if !bvm.ReadOnly {
+			mountOptions = append(mountOptions, "rw")
+		} else {
+			mountOptions = append(mountOptions, "rw")
+		}
```
adambkaplan, line 423: Can get rid of this

pkg/build/builder/daemonless.go
```diff
@@ -311,118 +328,233 @@ func buildDaemonlessImage(sc types.SystemContext, store storage.Store, isolation
...
+
+		if err := mountsMap.append(TransientMount{
+			BuildVolumeMount: bvm,
+			MountPathSource:  filepath.Join(BuildVolumeMountPath, bvm.MountPath),
```
adambkaplan, line 427: Now I see what you are doing in openshift-controller-manager - you are laying the mounts out based on their destination paths. It helps alleviate the volume name collision problem. `filepath.Join` also strips the leading `/` if we are given an absolute file path, which is good. We should probably convert relative file paths to an absolute filepath within buildah's runtime environment.

pkg/build/builder/daemonless.go
```diff
@@ -311,118 +328,233 @@ func buildDaemonlessImage(sc types.SystemContext, store storage.Store, isolation
...
+	if tm.NoDev {
+		mountOptions = append(mountOptions, "nodev")
+	}
+	if tm.NoExec {
+		mountOptions = append(mountOptions, "noexec")
+	}
+	if tm.NoSuid {
+		mountOptions = append(mountOptions, "nosuid")
+	}
+	return fmt.Sprintf("%s:%s:%s", tm.MountPathSource, tm.MountPath, strings.Join(mountOptions, ","))
```
adambkaplan, line 358: See my note about `MountPath` being relative - we should convert this to an absolute path.

gabemontero:
FYI there is auto linking of PRs to Jira items if you 

/retitle BUILD-87 secret configmap volume mounts in builds

gabemontero:
/retitle BUILD-87: secret configmap volume mounts in builds

gabemontero:
Course I don't have permissions to do it @coreydaley :-)

adambkaplan:
/hold

Given the complexity of the feature and the time remaining before 4.8 reaches "final freeze" period, we are going to hold this until master opens for 4.9

coreydaley:
@adambkaplan ptal

xiuwang:
Test on PR launched cluster launch openshift/builder#245,openshift/openshift-controller-manager#183,openshift/openshift-apiserver#208,openshift/oc#843 following test cases https://url.corp.redhat.com/buildvolumes , no critical issues found.

/label qe-approved

coreydaley:
@adambkaplan @gabemontero ptal, tests are passing and it is ready for review.

pkg/build/builder/transient_mounts.go
```diff
@@ -0,0 +1,106 @@
...
+
+}
+
+// TransientMounts is a map of transient mounts with the destination as the key
+// so we can detect duplicates quickly as we append new ones
+type TransientMounts map[string]TransientMount
+
+// appends a transient mount to the map and returns an error if a duplicate
+// destination is detected
+func (t TransientMounts) append(mount TransientMount) error {
```
gabemontero, line 60: might be the first time I've seen in practice a non-struct ptr used for this .... cool :-)

pkg/build/builder/transient_mounts.go
```diff
@@ -0,0 +1,106 @@
...
+//TransientMount is a structure to hold the data needed to construct a transient mount
+type TransientMount struct {
+	Source      string
+	Destination string
+	Options     TransientMountOptions
+}
+
+// returns the string representation of a transient mount
+// format: source:destiation:option1,option2,...
+func (tm TransientMount) String() string {
```
gabemontero, line 25: don't we typically employ `tm *TransientMount)` here @coreydaley ?

if there is a specific reason why you did not do it that way, I'm missing it (and perhaps a clarifying comment wrt that is needed)

pkg/build/builder/transient_mounts.go
gabemontero: btw, the map ref is a ptr under the cover iirc correctly, hence you need nil checks when checking map variables, re: related to my comment above with the funcs off of the `TransientMount` struct

pkg/build/builder/util.go
```diff
@@ -150,3 +154,15 @@ func ParseProxyURL(proxy string) (*url.URL, error) {
 
 	return proxyURL, err
 }
+
+// NameForBuildVolume returns a valid pod volume name for the provided build volume name.
```
gabemontero, line 158: Did I see an exchange in a library-go PR that landed with *NOT* including a method like this there?

I ask because I presume (though I have not visited your PR there super recently) there is a need for such a method in openshift-controller-manager, and we'd ideally share the impl (albeit this is a simple thing).

But it may not be a given, and if so, say as much @coreydaley and we can resolve this thread ... but I'll mention it here anyway just in case.

So just let me know and we can go from there.

gabemontero: we do have https://github.com/openshift/runtime-utils already for sharing build specific logic that could be a landing spot for such common methods that is specific to builds vs. library-go

pkg/build/builder/util.go
```diff
@@ -150,3 +154,15 @@ func ParseProxyURL(proxy string) (*url.URL, error) {
...
+func NameForBuildVolume(objName string) string {
+	// Volume names must be a valid DNS Label - see https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names
+	return naming.GetName(strings.ToLower(objName), buildVolumeSuffix, kvalidation.DNS1123LabelMaxLength)
+}
+
+// PathForBuildVolume returns the path in the builder container where the build volume is mounted.
```
gabemontero, line 164: same thing here wrt shared functions between builder and openshift-controller-manager

gabemontero (commented):
A few comments, but no big blockers ... other than perhaps sharing some functions between builder and OCM in https://github.com/openshift/runtime-utils most likely these are just questions of mine you can answer @coreydaley and we can move on.

pkg/build/builder/util.go
coreydaley: It was decided by the powers that be that this code should be duplicated between the repos instead of shared via library-go, and as such, the library-go pull request has been closed.

As for `runtime-utils`, all I see there is a `doc.go` in the build directory?

pkg/build/builder/util.go
coreydaley: It was decided by the powers that be that this code should be duplicated between the repos instead of shared via library-go, and as such, the library-go pull request has been closed.

pkg/build/builder/transient_mounts.go
coreydaley: Since this method is not mutating the underlying data structure the pointer type is not needed. I'm not sure that a comment for that is needed.

https://gobyexample.com/methods

pkg/build/builder/transient_mounts.go
coreydaley: > might be the first time I've seen in practice a non-struct ptr used for this .... cool :-)

I'm not going to say that this is common practice, or even the correct thing to do, but it worked :)

pkg/build/builder/util.go
gabemontero: > It was decided by the powers that be that this code should be duplicated between the repos instead of shared via library-go, and as such, the library-go pull request has been closed.
> 

ok so my recollection was correct in this case

> As for `runtime-utils`, all I see there is a `doc.go` in the build directory?

no there is also a pkg directory, with currently a "registries" subdir

technically speaking, per the readme, this repo was a common ground for https://github.com/containers related shared libs that build code would leverage, but there is no reason IMO would could not expand that a bit

@adambkaplan and @bparees are in the OWNER file

also note, that repo arose when similar utility functions were rejected by library-go

pkg/build/builder/transient_mounts.go
coreydaley: > btw, the map ref is a ptr under the cover iirc correctly, hence you need nil checks when checking map variables, re: related to my comment above with the funcs off of the `TransientMount` struct

I believe that `ok := t[mount.Destination];` equates to a `nil` check, as it returns `false` if the key does not exist. Beyond that we are really only doing a few operations here:
 - Checking for duplicate keys
 - Adding a key/value pair
 - Outputting the map as a slice of strings constructed from the transient mount that is the value
If there is a specific place you think that I should be checking for `nil` let me know and I will add it.

gabemontero: OK, then I'll make an official "nit" req to change it to a pointer ref

adambkaplan (approved):
/approve

@rolfedh @RickJWagner ptal - we need "docs-approved" and "px-approved" labels before we merge.

pkg/build/builder/transient_mounts.go
gabemontero: I don't think this one should be marked resolved yet @coreydaley

pkg/build/builder/transient_mounts.go
gabemontero: yeah I'm good with this one ... go ahead and mark this one resolved ... but the other thread needs to be "unresolved"

pkg/build/builder/transient_mounts.go
coreydaley: updated in current push

gabemontero (commented):
ok, employing the same 

/lgtm

with 

/hold

while we wait on @adambkaplan to chime in on expanding https://github.com/openshift/runtime-utils to capture these common build specific constants and funcs that are denied inclusion into openshift/library-go

if we don't use it, unhold @coreydaley

adambkaplan:
Keeping the hold for docs/px approvals.

I don't think we should expand the scope of runtime-utils to add this copied code, since it pollutes that repo's purpose. I am disappointed that our PR to library-go was rejected, since other code in that repository doesn't meet their criteria.

PR #246, opened 2021-05-27
BUILD-278: Check cgroup v1 and cgroup v2 Files for Quota

gabemontero:
/assign @adambkaplan 
/assign @nalind 

there was no existing unit testing around `GetCGroupLimits()` or `readIn64` that I saw

the simplicity of the fix did not motivate me to take that on, but of course am curious on your respective takes

### DIFF
--- a/pkg/build/builder/dockerutil.go
+++ b/pkg/build/builder/dockerutil.go
@@ -102,16 +102,23 @@ func tagImage(dockerClient DockerClient, image, name string) error {
 	})
 }
 
-// readInt64 reads a file containing a 64 bit integer value
+// readMaxStringOrInt64 reads a file containing a 64 bit integer value, or string "max",
 // and returns the value as an int64.  If the file contains
-// a value larger than an int64, it returns MaxInt64,
+// a value larger than an int64, or the string "max", it returns MaxInt64,
 // if the value is smaller than an int64, it returns MinInt64.
-func readInt64(filePath string) (int64, error) {
+func readMaxStringOrInt64(filePath string) (int64, error) {
 	data, err := ioutil.ReadFile(filePath)
 	if err != nil {
 		return -1, err
 	}
 	s := strings.TrimSpace(string(data))
+	if s == "max" {
+		return math.MaxInt64, nil
+	}
+	return parseInt64(s)
+}
+
+func parseInt64(s string) (int64, error) {
 	val, err := strconv.ParseInt(s, 10, 64)
 	// overflow errors are ok, we'll get return a math.MaxInt64 value which is more
 	// than enough anyway.  For underflow we'll return MinInt64 and the error.
--- a/pkg/build/builder/dockerutil_test.go
+++ b/pkg/build/builder/dockerutil_test.go
@@ -1,6 +1,10 @@
 package builder
 
 import (
+	"io/ioutil"
+	"math"
+	"os"
+	"path/filepath"
 	"reflect"
 	"testing"
 
@@ -211,6 +215,63 @@ func TestCGroupParentExtraction(t *testing.T) {
 		}
 		if parent != tc.expect {
 			t.Errorf("[%s] expected cgroup parent= %s, got %s", tc.name, tc.expect, parent)
+		}
+	}
+}
+
+func TestReadMaxStringOrInt64(t *testing.T) {
+	tests := []struct {
+		name        string
+		fileContent string
+		expectedVal int64
+		expectedErr bool
+	}{
+		{
+			name:        "bad-data",
+			fileContent: "wqertwert",
+			expectedErr: true,
+		},
+		{
+			name:        "use-of-max",
+			fileContent: "max",
+			expectedVal: math.MaxInt64,
+		},
+		{
+			name:        "normal-int",
+			fileContent: "1234567891011",
+			expectedVal: int64(1234567891011),
+		},
+		{
+			name:        "file-missing",
+			expectedErr: true,
+		},
+	}
+	tmpDir, err := ioutil.TempDir(os.TempDir(), t.Name())
+	if err != nil {
+		t.Fatalf("error creating tmp dir: %s", err.Error())
+	}
+	defer os.RemoveAll(tmpDir)
+	for _, tc := range tests {
+		t.Logf("running tc %s", tc.name)
+		fileName := filepath.Join(tmpDir, tc.name)
+		if len(tc.fileContent) > 0 {
+			err = ioutil.WriteFile(fileName, []byte(tc.fileContent), 0644)
+			if err != nil {
+				t.Errorf("error writing data to file %s: %s", fileName, err.Error())
+			}
+		}
+		val, err := readMaxStringOrInt64(fileName)
+		if tc.expectedErr {
+			if err == nil {
+				t.Errorf("test %s expected error and did not get one", tc.name)
+			}
+			continue
+		}
+		if err != nil {
+			t.Errorf("test %s did not expect error and got: %s", tc.name, err.Error())
+		}
+		if tc.expectedVal != val {
+			t.Errorf("test %s expected val %v and got %v", tc.name, tc.expectedVal, val)
 		}
 	}
 }
--- a/pkg/build/builder/util_linux.go
+++ b/pkg/build/builder/util_linux.go
@@ -5,7 +5,6 @@ import (
 	"os"
 
 	"github.com/opencontainers/runc/libcontainer/cgroups"
-
 	s2iapi "github.com/openshift/source-to-image/pkg/api"
 )
 
@@ -13,7 +12,15 @@ import (
 // from the local /sys/fs/cgroup filesystem.  Overflow values are set to
 // math.MaxInt64.
 func GetCGroupLimits() (*s2iapi.CGroupLimits, error) {
-	byteLimit, err := readInt64("/sys/fs/cgroup/memory/memory.limit_in_bytes")
+	// see https://git.kernel.org/pub/scm/linux/kernel/git/tj/cgroup.git/tree/Documentation/admin-guide/cgroup-v2.rst
+	// for list of cgroupv2 files to try, but Nalin relayed that examination of the crun and runc code that 'memory.high'
+	// is not used.
+	file := "/sys/fs/cgroup/memory/memory.limit_in_bytes"
+	if cgroups.IsCgroup2UnifiedMode() {
+		file = "/sys/fs/cgroup/memory.max"
+	}
+	byteLimit, err := readMaxStringOrInt64(file)
+
 	if err != nil {
 		// for systems without cgroups builds should succeed
 		if _, err := os.Stat("/sys/fs/cgroup"); os.IsNotExist(err) {


pkg/build/builder/util_linux.go
```diff
@@ -13,7 +12,21 @@ import (
 // from the local /sys/fs/cgroup filesystem.  Overflow values are set to
 // math.MaxInt64.
 func GetCGroupLimits() (*s2iapi.CGroupLimits, error) {
-	byteLimit, err := readInt64("/sys/fs/cgroup/memory/memory.limit_in_bytes")
+	// see https://git.kernel.org/pub/scm/linux/kernel/git/tj/cgroup.git/tree/Documentation/admin-guide/cgroup-v2.rst
+	// for list of cgroupv2 files to try ... both support 'max' as a value in addition to int64
+	filesToTry := []string{"/sys/fs/cgroup/memory/memory.limit_in_bytes",
+		"/sys/fs/cgroup/memory.high",
+		" /sys/fs/cgroup/memory.max",
```
nalind, line 19:
```suggestion
		"/sys/fs/cgroup/memory.max",
```

pkg/build/builder/util_linux.go
```diff
@@ -13,7 +12,21 @@ import (
 // from the local /sys/fs/cgroup filesystem.  Overflow values are set to
 // math.MaxInt64.
 func GetCGroupLimits() (*s2iapi.CGroupLimits, error) {
-	byteLimit, err := readInt64("/sys/fs/cgroup/memory/memory.limit_in_bytes")
+	// see https://git.kernel.org/pub/scm/linux/kernel/git/tj/cgroup.git/tree/Documentation/admin-guide/cgroup-v2.rst
+	// for list of cgroupv2 files to try ... both support 'max' as a value in addition to int64
+	filesToTry := []string{"/sys/fs/cgroup/memory/memory.limit_in_bytes",
+		"/sys/fs/cgroup/memory.high",
```
nalind, line 18: A quick spin through the crun and runc sources suggests that neither of them set `memory.high`, so we probably want to skip it.

pkg/build/builder/util_linux.go
```diff
@@ -2,9 +2,8 @@ package builder
 
 import (
 	"fmt"
-	"os"
-
 	"github.com/opencontainers/runc/libcontainer/cgroups"
+	"os"
```
nalind, line 6: Probably want to drop this hunk.

pkg/build/builder/util_linux.go
gabemontero: update thanks

pkg/build/builder/util_linux.go
gabemontero: moved thanks

pkg/build/builder/util_linux.go
gabemontero: ok, which lines up with Guiseppe just suggesting `memory.max` in the bugzilla

gabemontero:
I think that is all your comments @nalind thanks

pkg/build/builder/dockerutil.go
```diff
@@ -126,6 +126,22 @@ func readInt64(filePath string) (int64, error) {
 	return val, nil
 }
 
+func readMaxStringOrInt64(filePath string) (int64, error) {
```
adambkaplan, line 129: Need a unit test for this function

adambkaplan (changes requested):

pkg/build/builder/dockerutil.go
gabemontero: so per my comment in https://github.com/openshift/builder/pull/246#issue-655174072 it would make sense to add unit tests not just for this function, but some of the existing code which does not have unit tests (i.e. `readInt64`)

so you are officially asking for that @adambkaplan ?

adambkaplan: Yes - there is a general ask across OpenShift to improve test code coverage. Adding scoped unit tests to new functions (and in this case, new functions that extend existing functions) moves the needle in the right direction.

gabemontero: unit test added @adambkaplan

gabemontero: actually sit tight @adambkaplan 

per https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_builder/246/pull-ci-openshift-builder-master-unit/1397983052998643712#1:build-log.txt%3A12

may have a golang version discrepancy to sort out .... `os.WriteFile` works for me locally but is noted as undefined in CI

gabemontero: sure  enough, things moved between golang 1.15 and 1.16 per the overview in https://golang.org/pkg/io/ioutil/ and openshift/builder is still at 1.15 per https://github.com/openshift/release/blob/master/ci-operator/config/openshift/builder/openshift-builder-master.yaml#L37

I know you have limited availability this afternoon @adambkaplan so I'm for now going to assume we do not want to bump golang for openshift/builder this late in the cycle, and I'll switch from os.WriteFile to io/ioutil.WriteFile

gabemontero: ok unit tests now clean @adambkaplan

pkg/build/builder/dockerutil.go
```diff
@@ -126,6 +126,27 @@ func readInt64(filePath string) (int64, error) {
...
+	val, err := readInt64(filePath)
+	if err == nil {
+		return val, nil
+	}
+	if os.IsNotExist(err) {
+		return val, err
+	}
+	// see if we have a string called 'max' for the value
+	data, err2 := ioutil.ReadFile(filePath)
```
kolyshkin, line 138: So basically you're reading the file twice -- this is redundant.

I suggest to separate out the portion of readInt64 that parses the value and re-use it from here.

Something like this maybe

```diff
diff --git a/pkg/build/builder/dockerutil.go b/pkg/build/builder/dockerutil.go
index 764e06655..1ef30681a 100644
--- a/pkg/build/builder/dockerutil.go
+++ b/pkg/build/builder/dockerutil.go
@@ -111,7 +111,24 @@ func readInt64(filePath string) (int64, error) {
        if err != nil {
                return -1, err
        }
+
+       return parseInt64(strings.TrimSpace(string(data)))
+}
+
+func readMaxStringOrInt64(filePath string) (int64, error) {
+       data, err := ioutil.ReadFile(filePath)
+       if err != nil {
+               return -1, err
+       }
        s := strings.TrimSpace(string(data))
+       if s == "max" {
+               return math.MaxInt64, nil
+       }
+
+       return parseInt64(s)
+}
+
+func parseInt64(s string) (int64, error) {
        val, err := strconv.ParseInt(s, 10, 64)
        // overflow errors are ok, we'll get return a math.MaxInt64 value which is more
        // than enough anyway.  For underflow we'll return MinInt64 and the error.
```

pkg/build/builder/dockerutil_test.go
```diff
@@ -214,3 +218,69 @@ func TestCGroupParentExtraction(t *testing.T) {
 		}
 	}
 }
+
+func writeFile(fileName, fileContent string, t *testing.T) {
+	file, err := os.Create(fileName)
+	if err != nil {
+		t.Errorf("error creating file %s: %s", fileName, err.Error())
+	}
+	defer file.Close()
```
kolyshkin, line 227: No need for this as `ioutil.WriteFile` creates the file.

pkg/build/builder/dockerutil_test.go
```diff
@@ -214,3 +218,69 @@ func TestCGroupParentExtraction(t *testing.T) {
...
+	err = ioutil.WriteFile(fileName, []byte(fileContent), 0777)
```
kolyshkin, line 228: `0777` looks a bit excessive. Perhaps
```suggestion
	err = ioutil.WriteFile(fileName, []byte(fileContent), 0o644)
```

pkg/build/builder/util_linux.go
```diff
@@ -5,15 +5,28 @@ import (
...
 
 // GetCGroupLimits returns a struct populated with cgroup limit values gathered
 // from the local /sys/fs/cgroup filesystem.  Overflow values are set to
 // math.MaxInt64.
 func GetCGroupLimits() (*s2iapi.CGroupLimits, error) {
-	byteLimit, err := readInt64("/sys/fs/cgroup/memory/memory.limit_in_bytes")
+	// see https://git.kernel.org/pub/scm/linux/kernel/git/tj/cgroup.git/tree/Documentation/admin-guide/cgroup-v2.rst
+	// for list of cgroupv2 files to try, but Nalin relayed that examination of the crun and runc code that 'memory.high'
+	// is not used.
+	filesToTry := []string{"/sys/fs/cgroup/memory/memory.limit_in_bytes",
```
kolyshkin, line 18: Formatting nit:
```suggestion
	filesToTry := []string{
		"/sys/fs/cgroup/memory/memory.limit_in_bytes",
```

pkg/build/builder/util_linux.go
```diff
@@ -5,15 +5,28 @@ import (
...
+		"/sys/fs/cgroup/memory.max",
+	}
+	var byteLimit int64
+	var err error
+	for _, filePath := range filesToTry {
```
kolyshkin, line 23: Trying to read cgroup v2 file on a cgroup v1 system, or vice versa, is not very efficient.

Trying to read cgroup v2 file on _any_ error from reading cgroup v1 file is also not good.

I suggest using something like IsCgroupV2UnifiedMode() from https://github.com/opencontainers/runc/blob/960b5fdb5acde1063960a80eb224672faf244e02/libcontainer/cgroups/utils.go#L34

and do this:
```go
file := "/sys/fs/cgroup/memory/memory.limit_in_bytes"
if IsCgroupV2UnifiedMode() {
   file = "/sys/fs/cgroup/memory.max"
}
byteLimit, err := readMaxStringOrInt64(file)
....
```

kolyshkin (changes requested):

gabemontero:
updates pushed @kolyshkin @nalind ptal thanks

pkg/build/builder/dockerutil.go
```diff
@@ -124,6 +131,7 @@ func readInt64(filePath string) (int64, error) {
 		return -1, err
 	}
 	return val, nil
+
```
kolyshkin, line 134: nit: extra whitespace

kolyshkin: Funny there's an (very old) open issue for gofmt to address this particular case (https://github.com/golang/go/issues/6996)

kolyshkin (commented):
a whitespace nit, otherwise LGTM

adambkaplan (approved):
/approve

Outstanding format nit, otherwise looks good.

gabemontero:
white space removed @adambkaplan @kolyshkin thanks folks

pkg/build/builder/util_linux.go
```diff
@@ -5,15 +5,22 @@ import (
...
 )
 
 // GetCGroupLimits returns a struct populated with cgroup limit values gathered
 // from the local /sys/fs/cgroup filesystem.  Overflow values are set to
 // math.MaxInt64.
 func GetCGroupLimits() (*s2iapi.CGroupLimits, error) {
-	byteLimit, err := readInt64("/sys/fs/cgroup/memory/memory.limit_in_bytes")
+	// see https://git.kernel.org/pub/scm/linux/kernel/git/tj/cgroup.git/tree/Documentation/admin-guide/cgroup-v2.rst
+	// for list of cgroupv2 files to try, but Nalin relayed that examination of the crun and runc code that 'memory.high'
+	// is not used.
```
kolyshkin, line 17: Technically this comment belongs to cgroupv2 case inside the `if` block, but that's not important.

In the future, though, we might need to also check `memory.high` (in case `memory.max` returns `max`.

kolyshkin (approved):
LGTM

pkg/build/builder/util_linux.go
```diff
@@ -5,15 +5,22 @@ import (
...
+	file := "/sys/fs/cgroup/memory/memory.limit_in_bytes"
+	if cgroups.IsCgroup2UnifiedMode() {
+		file = "/sys/fs/cgroup/memory.max"
```
kolyshkin, line 20: Sorry I haven't raised this concern earlier, but here it is.

cgroup v2 does not have "memory.max" file in the top-level cgroup **on the host**, so this won't work **on the host** (unless you add a fallback to assume the limit is `max` or somesuch).

If this is done inside a container, this check will work, as the container has its own cgroup mounted to /sys/fs/cgroup, not the top one.

Hope that helps.

gabemontero: thanks @kolyshkin - that clarification, and the additional host level verification details you added in the associated bugzilla, should help with the confusion @wewang58 was having in trying to verify a proper set up to verify.

As it turns out, this logic is in fact running in a container running in the build pod.

@adambkaplan I think we are ready for the "official" lgtm so we can get this merged, unless you had any other questions.

gabemontero:
per @rphillips cgroupsv2 support is out wrt openshift and it is "too soon" to drop this in.... QE has now "user acceptable" form for installing openshift with cgroupsv2

I'm going to

/hold 

this vs. closing for now to make it easier to find and resuscitate this when in fact cgroupsv2 gets officially into openshift plan pipeline

thanks again @kolyshkin for you assistance

adambkaplan:
/retitle BUILD-278: Check cgroup v1 and cgroup v2 Files for Quota

gabemontero:
@vrutkovs mentioned in bz 1949438 that at the moment cgroupsv2 is tech preview for 4.9

pkg/build/builder/util_linux.go
kolyshkin: It's all good then.

kolyshkin (approved):
LGTM

vrutkovs:
>CI is not ready for it yet 

CI job to test this is added in https://github.com/openshift/release/pull/19115, I suppose this is good to go in 4.9 now?

pkg/build/builder/util_linux.go
giuseppe: for completeness, both runc/crun supports setting `memory.high` through the unified map (podman exposes it with `--cgroup-conf`) but I don't think we should worry about it, as a cgroup is still allowed to allocate more memory until it reaches `memory.max`

giuseppe (approved):
LGTM

adambkaplan (approved):
/lgtm

adambkaplan:
/hold cancel

I imagine we need to merge this for the cgroupsv2 jobs to go green.

gabemontero:
> /hold cancel
> 
> I imagine we need to merge this for the cgroupsv2 jobs to go green.

last I heard @adambkaplan about this was https://github.com/openshift/builder/pull/246#issuecomment-853860972 and a ping from @rphillips where he said CI was not ready for cgroupsv2

just double checking are you aware with master branch opening up for 4.9 where the install changes needed for CI (and test for that matter) to get this are in place?

adambkaplan:
> just double checking are you aware with master branch opening up for 4.9 where the install changes needed for CI (and test for that matter) to get this are in place?

Yes - ack that we won't have CI running this end to end until we can install with cgroupsv2.

gabemontero:
Hey @kolyshkin @nalind : so @vrutkovs is trying some cgroupv2 CI rehearsals over with PR 

In that PR, where presumably cgroupv2 is set up correctly, I can confirm in the logs that our us our use of the openshicontainers/runc API is kicking in at 

https://github.com/openshift/builder/blob/814bbb090228374742e31889c7fe3df0c3662c7a/pkg/build/builder/util_linux.go#L19-L21

because I see the code picking `/sys/fs/cgroup/memory.max` as the file to access.

However, we are ultimately getting error logs like 

`2021-07-12T09:05:29.940355077Z error: failed to retrieve cgroup limits: cannot determine cgroup limits: open /sys/fs/cgroup/memory.max: no such file or directory`

Did we get the name of the file wrong?
We currently do not assume the file is there with https://github.com/openshift/builder/blob/814bbb090228374742e31889c7fe3df0c3662c7a/pkg/build/builder/util_linux.go#L26-L28 but perhaps seeing if `/sys/fs/cgroup` not existing is an insufficient check for ignoring the lack of `/sys/fs/cgroup/memory.max` ?

Or do you see something in the test setup @vrutkovs has ?

I could start adding (short term/temporary) debug in that path to display the full contents of `/sys/fs.cgroup` but I decided to pull you guys in first and get any possibly quick sanity checks from you all.

@adambkaplan fyi

thanks

gabemontero:
edited my prior comment with the link to @vrutkovs 's test enablement PR

vrutkovs:
Here's what I see on the node:
```
$ oc debug node/ip-10-0-140-154.ec2.internal                                                      
Starting pod/ip-10-0-140-154ec2internal-debug ...
To use host binaries, run `chroot /host`
chroot /host
Pod IP: 10.0.140.154
If you don't see a command prompt, try pressing enter.
sh-4.4# chroot /host
sh-4.4# cat /proc/cmdline 
BOOT_IMAGE=(hd0,gpt3)/ostree/rhcos-20246fb86657495594dfd90b85bcfa58f8a400916e1580611a5b7ea95e77f783/vmlinuz-4.18.0-305.7.1.el8_4.x86_64 random.trust_cpu=on console=tty0 console=ttyS0,115200n8 ostree=/ostree/boot.0/rhcos/20246fb86657495594dfd90b85bcfa58f8a400916e1580611a5b7ea95e77f783/0 ignition.platform.id=aws root=UUID=12afac90-6b11-458a-9f0b-1f5303f2b1f9 rw rootflags=prjquota systemd.unified_cgroup_hierarchy=1
sh-4.4# ls -la /sys/fs/cgroup/
total 0
dr-xr-xr-x.  6 root root 0 Jul 12 15:51 .
drwxr-xr-x.  9 root root 0 Jul 12 15:50 ..
-r--r--r--.  1 root root 0 Jul 12 15:50 cgroup.controllers
-rw-r--r--.  1 root root 0 Jul 12 15:51 cgroup.max.depth
-rw-r--r--.  1 root root 0 Jul 12 15:51 cgroup.max.descendants
-rw-r--r--.  1 root root 0 Jul 12 15:50 cgroup.procs
-r--r--r--.  1 root root 0 Jul 12 15:51 cgroup.stat
-rw-r--r--.  1 root root 0 Jul 12 16:12 cgroup.subtree_control
-rw-r--r--.  1 root root 0 Jul 12 15:51 cgroup.threads
-rw-r--r--.  1 root root 0 Jul 12 15:51 cpu.pressure
-r--r--r--.  1 root root 0 Jul 12 15:51 cpuset.cpus.effective
-r--r--r--.  1 root root 0 Jul 12 15:51 cpuset.mems.effective
drwxr-xr-x.  2 root root 0 Jul 12 15:50 init.scope
-rw-r--r--.  1 root root 0 Jul 12 15:51 io.pressure
-r--r--r--.  1 root root 0 Jul 12 15:51 io.stat
drwxr-xr-x.  4 root root 0 Jul 12 15:51 kubepods.slice
-rw-r--r--.  1 root root 0 Jul 12 15:51 memory.pressure
-r--r--r--.  1 root root 0 Jul 12 15:51 memory.stat
drwxr-xr-x. 77 root root 0 Jul 12 15:50 system.slice
drwxr-xr-x.  2 root root 0 Jul 12 15:51 user.slice
```

And in the container:
```
$ oc run fedora --image=quay.io/fedora/fedora:34 --command -- sleep infinity
pod/fedora created
$ oc exec -ti fedora -- sh
sh-5.0# ls -la /sys/fs/cgroup/
total 0
drwxr-xr-x.  2 root root 0 Jul 12 16:15 .
drwxr-xr-x. 10 root root 0 Jul 12 16:15 ..
-r--r--r--.  1 root root 0 Jul 12 16:15 cgroup.controllers
-r--r--r--.  1 root root 0 Jul 12 16:15 cgroup.events
-rw-r--r--.  1 root root 0 Jul 12 16:15 cgroup.freeze
-rw-r--r--.  1 root root 0 Jul 12 16:15 cgroup.max.depth
-rw-r--r--.  1 root root 0 Jul 12 16:15 cgroup.max.descendants
-rw-r--r--.  1 root root 0 Jul 12 16:15 cgroup.procs
-r--r--r--.  1 root root 0 Jul 12 16:15 cgroup.stat
-rw-r--r--.  1 root root 0 Jul 12 16:15 cgroup.subtree_control
-rw-r--r--.  1 root root 0 Jul 12 16:15 cgroup.threads
-rw-r--r--.  1 root root 0 Jul 12 16:15 cgroup.type
-rw-r--r--.  1 root root 0 Jul 12 16:15 cpu.max
-rw-r--r--.  1 root root 0 Jul 12 16:15 cpu.pressure
-r--r--r--.  1 root root 0 Jul 12 16:15 cpu.stat
-rw-r--r--.  1 root root 0 Jul 12 16:15 cpu.weight
-rw-r--r--.  1 root root 0 Jul 12 16:15 cpu.weight.nice
-rw-r--r--.  1 root root 0 Jul 12 16:15 cpuset.cpus
-r--r--r--.  1 root root 0 Jul 12 16:15 cpuset.cpus.effective
-rw-r--r--.  1 root root 0 Jul 12 16:15 cpuset.cpus.partition
-rw-r--r--.  1 root root 0 Jul 12 16:15 cpuset.mems
-r--r--r--.  1 root root 0 Jul 12 16:15 cpuset.mems.effective
-r--r--r--.  1 root root 0 Jul 12 16:15 hugetlb.1GB.current
-r--r--r--.  1 root root 0 Jul 12 16:15 hugetlb.1GB.events
-r--r--r--.  1 root root 0 Jul 12 16:15 hugetlb.1GB.events.local
-rw-r--r--.  1 root root 0 Jul 12 16:15 hugetlb.1GB.max
-r--r--r--.  1 root root 0 Jul 12 16:15 hugetlb.2MB.current
-r--r--r--.  1 root root 0 Jul 12 16:15 hugetlb.2MB.events
-r--r--r--.  1 root root 0 Jul 12 16:15 hugetlb.2MB.events.local
-rw-r--r--.  1 root root 0 Jul 12 16:15 hugetlb.2MB.max
-rw-r--r--.  1 root root 0 Jul 12 16:15 io.bfq.weight
-rw-r--r--.  1 root root 0 Jul 12 16:15 io.latency
-rw-r--r--.  1 root root 0 Jul 12 16:15 io.max
-rw-r--r--.  1 root root 0 Jul 12 16:15 io.pressure
-r--r--r--.  1 root root 0 Jul 12 16:15 io.stat
-r--r--r--.  1 root root 0 Jul 12 16:15 memory.current
-r--r--r--.  1 root root 0 Jul 12 16:15 memory.events
-rw-r--r--.  1 root root 0 Jul 12 16:15 memory.high
-rw-r--r--.  1 root root 0 Jul 12 16:15 memory.low
-rw-r--r--.  1 root root 0 Jul 12 16:15 memory.max
-rw-r--r--.  1 root root 0 Jul 12 16:15 memory.min
-rw-r--r--.  1 root root 0 Jul 12 16:15 memory.oom.group
-rw-r--r--.  1 root root 0 Jul 12 16:15 memory.pressure
-r--r--r--.  1 root root 0 Jul 12 16:15 memory.stat
-r--r--r--.  1 root root 0 Jul 12 16:15 memory.swap.current
-r--r--r--.  1 root root 0 Jul 12 16:15 memory.swap.events
-rw-r--r--.  1 root root 0 Jul 12 16:15 memory.swap.high
-rw-r--r--.  1 root root 0 Jul 12 16:15 memory.swap.max
-r--r--r--.  1 root root 0 Jul 12 16:15 pids.current
-r--r--r--.  1 root root 0 Jul 12 16:15 pids.events
-rw-r--r--.  1 root root 0 Jul 12 16:15 pids.max
-r--r--r--.  1 root root 0 Jul 12 16:15 rdma.current
-rw-r--r--.  1 root root 0 Jul 12 16:15 rdma.max
```

gabemontero:
Thanks @vrutkovs - certainly helpful

So the current call for cgroup stuff is in the build **pod**, when it is prepping the build container, which is potentially different than both your node snapshot (where `memory.max` is missing) and your direct fedora container invocation snaphsot (where `memory.max` is present).

Certainly in both, `/sys/fs/cgroup/` exists.

The reasonable assumption based on the behavior exhibited would be that the build pod looks like the node.  A dir structure like the node would result in the behavior exhibited.

My sense is that we should still verify the contents while in the build pod.  But assuming it is like the node, then get agreement when `/sys/fs/cgroup` exists, but `memory.max` does not, we assume that this the "default" case.

But I do not consider myself an SME here and would like some consensus / confirmation from @kolyshkin @nalind @adambkaplan

gabemontero:
I've created https://github.com/openshift/builder/pull/252 to potentially address the results of @vrutkovs 's CI testing attempt and move the associated discussion to a open vs. merged PR

PR #252, opened 2021-07-12
BUILD-278: fix cgroupv2 memory max defaulting

gabemontero:
See results from https://github.com/openshift/builder/pull/246#issuecomment-878357189 and later discussion along with https://github.com/openshift/release/pull/19115

This is an attempt to adjust and address the various concerns.

/assign @kolyshkin 
/assign @nalind 
/assign @adambkaplan 

@vrutkovs FYI


gabemontero:
/hold 

@vrutkovs has https://github.com/openshift/release/pull/20125 up so that we can test these cgroupv2 changes via a new optional e2e job (thanks @vrutkovs !!)

once that merges, we'll drive test of this change from that

vrutkovs:
again
>error: failed to retrieve cgroup limits: cannot determine cgroup limits: open /sys/fs/cgroup/memory.max: no such file or directory

in the build logs

gabemontero:
> again
> 
> > error: failed to retrieve cgroup limits: cannot determine cgroup limits: open /sys/fs/cgroup/memory.max: no such file or directory
> 
> in the build logs

yep

I'll start adding debug prints along the various paths here in this PR and we'll go from there.

gabemontero:
I may have figured out what was wrong with the patch here while adding debug.  Will add that change in addition to the debug.  Should be up shortly.

gabemontero:
AWS VpcLimitExceeded error on last e2e-aws-builds failure

will retest after the other e2e's come in just in case this is widespread

gabemontero:
Combo of incorrect debug and issues with the is dir present check.  Update should be uploaded momentarily.

```
Log Tail:	GGM cgroup2 true
		GGM file /sys/fs/cgroup/memory.max go err open /sys/fs/cgroup/memory.max: no such file or directory
		GGM stat on dir got err &fs.PathError{Op:"open", Path:"/sys/fs/cgroup/memory.max", Err:0x2}
		GGM notExist false
		error: failed to retrieve cgroup limits: read cgroup paren...d498b2b2c61f86213fd5f42e3293f49f1298185122f58e994855.scope]

```

gabemontero:
Another test had a more thorough dump on the error.  We are not in the no memory.max provided, default path, but get subsequent errors.  Investigating:

```
2021-07-14T13:45:46.844072193Z I0714 13:45:46.844054       1 util_linux.go:57] found cgroup values map: map[:/kubepods.slice/kubepods-besteffort.slice/kubepods-besteffort-pod340d7682_7be8_4207_b567_23c490506337.slice/crio-e8bdd41c69d41f8060534462dd978b49ab44caa665022fa97e6ed2f9b935a310.scope]
2021-07-14T13:45:46.850317302Z F0714 13:45:46.850283       1 helpers.go:115] error: failed to retrieve cgroup limits: read cgroup parent: could not find memory cgroup subsystem in map map[:/kubepods.slice/kubepods-besteffort.slice/kubepods-besteffort-pod340d7682_7be8_4207_b567_23c490506337.slice/crio-e8bdd41c69d41f8060534462dd978b49ab44caa665022fa97e6ed2f9b935a310.scope]
2021-07-14T13:45:46.850481404Z goroutine 1 [running]:
2021-07-14T13:45:46.850481404Z k8s.io/klog/v2.stacks(0xc000012001, 0xc000a461c0, 0x157, 0x1a8)
2021-07-14T13:45:46.850481404Z 	/go/src/github.com/openshift/builder/vendor/k8s.io/klog/v2/klog.go:1021 +0xb9
2021-07-14T13:45:46.850481404Z k8s.io/klog/v2.(*loggingT).output(0x3c32c60, 0xc000000003, 0x0, 0x0, 0xc000036070, 0x302d4cf, 0xa, 0x73, 0xcb8800)
2021-07-14T13:45:46.850481404Z 	/go/src/github.com/openshift/builder/vendor/k8s.io/klog/v2/klog.go:970 +0x191
2021-07-14T13:45:46.850481404Z k8s.io/klog/v2.(*loggingT).printDepth(0x3c32c60, 0xc000000003, 0x0, 0x0, 0x0, 0x0, 0x2, 0xc0005f81c0, 0x1, 0x1)
2021-07-14T13:45:46.850481404Z 	/go/src/github.com/openshift/builder/vendor/k8s.io/klog/v2/klog.go:733 +0x16f
2021-07-14T13:45:46.850481404Z k8s.io/klog/v2.FatalDepth(...)
2021-07-14T13:45:46.850481404Z 	/go/src/github.com/openshift/builder/vendor/k8s.io/klog/v2/klog.go:1495
2021-07-14T13:45:46.850481404Z k8s.io/kubectl/pkg/cmd/util.fatal(0xc00022e8c0, 0x128, 0x1)
2021-07-14T13:45:46.850481404Z 	/go/src/github.com/openshift/builder/vendor/k8s.io/kubectl/pkg/cmd/util/helpers.go:93 +0x288
2021-07-14T13:45:46.850481404Z k8s.io/kubectl/pkg/cmd/util.checkErr(0x2c02e80, 0xc0004d5d80, 0x2a404e0)
2021-07-14T13:45:46.850481404Z 	/go/src/github.com/openshift/builder/vendor/k8s.io/kubectl/pkg/cmd/util/helpers.go:188 +0x935
2021-07-14T13:45:46.850481404Z k8s.io/kubectl/pkg/cmd/util.CheckErr(...)
2021-07-14T13:45:46.850481404Z 	/go/src/github.com/openshift/builder/vendor/k8s.io/kubectl/pkg/cmd/util/helpers.go:115
2021-07-14T13:45:46.850481404Z main.NewCommandDockerBuilder.func1(0xc000384000, 0xc00007e840, 0x0, 0x1)
2021-07-14T13:45:46.850481404Z 	/go/src/github.com/openshift/builder/cmd/builder.go:125 +0x129
2021-07-14T13:45:46.850481404Z main.main.func2(0xc000384000, 0xc00007e840, 0x0, 0x1)
2021-07-14T13:45:46.850481404Z 	/go/src/github.com/openshift/builder/cmd/main.go:85 +0x239
2021-07-14T13:45:46.850481404Z github.com/spf13/cobra.(*Command).execute(0xc000384000, 0xc000050070, 0x1, 0x1, 0xc000384000, 0xc000050070)
2021-07-14T13:45:46.850481404Z 	/go/src/github.com/openshift/builder/vendor/github.com/spf13/cobra/command.go:856 +0x2c2
2021-07-14T13:45:46.850481404Z github.com/spf13/cobra.(*Command).ExecuteC(0xc000384000, 0xc000520720, 0x2938405, 0xd)
2021-07-14T13:45:46.850481404Z 	/go/src/github.com/openshift/builder/vendor/github.com/spf13/cobra/command.go:960 +0x375
2021-07-14T13:45:46.850481404Z github.com/spf13/cobra.(*Command).Execute(...)
2021-07-14T13:45:46.850481404Z 	/go/src/github.com/openshift/builder/vendor/github.com/spf13/cobra/command.go:897
2021-07-14T13:45:46.850481404Z main.main()
2021-07-14T13:45:46.850481404Z 	/go/src/github.com/openshift/builder/cmd/main.go:88 +0xac5
2021-07-14T13:45:46.850481404Z 
2021-07-14T13:45:46.850481404Z goroutine 6 [chan receive]:
2021-07-14T13:45:46.850481404Z k8s.io/klog/v2.(*loggingT).flushDaemon(0x3c32c60)
2021-07-14T13:45:46.850481404Z 	/go/src/github.com/openshift/builder/vendor/k8s.io/klog/v2/klog.go:1164 +0x8b
2021-07-14T13:45:46.850481404Z created by k8s.io/klog/v2.init.0
2021-07-14T13:45:46.850481404Z 	/go/src/github.com/openshift/builder/vendor/k8s.io/klog/v2/klog.go:418 +0xdf
2021-07-14T13:45:46.850481404Z 
2021-07-14T13:45:46.850481404Z goroutine 10 [syscall]:
2021-07-14T13:45:46.850481404Z os/signal.signal_recv(0x0)
2021-07-14T13:45:46.850481404Z 	/usr/lib/golang/src/runtime/sigqueue.go:168 +0xa5
2021-07-14T13:45:46.850481404Z os/signal.loop()
2021-07-14T13:45:46.850481404Z 	/usr/lib/golang/src/os/signal/signal_unix.go:23 +0x25
2021-07-14T13:45:46.850481404Z created by os/signal.Notify.func1.1
2021-07-14T13:45:46.850481404Z 	/usr/lib/golang/src/os/signal/signal.go:151 +0x45
2021-07-14T13:45:46.850481404Z 
2021-07-14T13:45:46.850481404Z goroutine 11 [chan receive]:
2021-07-14T13:45:46.850481404Z main.main.func1(0xc000330d80)
2021-07-14T13:45:46.850481404Z 	/go/src/github.com/openshift/builder/cmd/main.go:34 +0x38
2021-07-14T13:45:46.850481404Z created by main.main
2021-07-14T13:45:46.850481404Z 	/go/src/github.com/openshift/builder/cmd/main.go:33 +0x14a
2021-07-14T13:45:46.850481404Z 
2021-07-14T13:45:46.850481404Z goroutine 12 [select]:
2021-07-14T13:45:46.850481404Z k8s.io/apimachinery/pkg/util/wait.BackoffUntil(0x2a404a8, 0x2c05a40, 0xc000576000, 0x1, 0xc000054c60)
2021-07-14T13:45:46.850481404Z 	/go/src/github.com/openshift/builder/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:167 +0x118
2021-07-14T13:45:46.850481404Z k8s.io/apimachinery/pkg/util/wait.JitterUntil(0x2a404a8, 0x12a05f200, 0x0, 0x1, 0xc000054c60)
2021-07-14T13:45:46.850481404Z 	/go/src/github.com/openshift/builder/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:133 +0x98
2021-07-14T13:45:46.850481404Z k8s.io/apimachinery/pkg/util/wait.Until(...)
2021-07-14T13:45:46.850481404Z 	/go/src/github.com/openshift/builder/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:90
2021-07-14T13:45:46.850481404Z k8s.io/apimachinery/pkg/util/wait.Forever(0x2a404a8, 0x12a05f200)
2021-07-14T13:45:46.850481404Z 	/go/src/github.com/openshift/builder/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:81 +0x4f
2021-07-14T13:45:46.850481404Z created by k8s.io/component-base/logs.InitLogs
2021-07-14T13:45:46.850481404Z 	/go/src/github.com/openshift/builder/vendor/k8s.io/component-base/logs/logs.go:58 +0x8a

```

gabemontero:
next tweak up wrt default when no `memory.max`

gabemontero:
VpcLimitExceeded with latest e2e-aws-image-ecosystem

gabemontero:
Down to one failure in e2e-aws-cgroupsv2 @vrutkovs 

: [sig-builds][Feature:Builds] s2i build with a quota Building from a template should create an s2i build with a quota and run it [Skipped:Disconnected] [Suite:openshift/conformance/parallel] expand_less

quota and cgroups2 certainly could have a relationship that is unique vs. the other test

investigating

gabemontero:
So I see this in the buildah debug @vrutkovs @nalind @adambkaplan 

```
    time="2021-07-14T16:33:45Z" level=debug msg="Running [\"runc\" \"start\" \"buildah-buildah101642493\"]"
    MEMORY=cat: /sys/fs/cgroup/memory/memory.limit_in_bytes: No such file or directory
    MEMORYSWAP=cat: /sys/fs/cgroup/memory/memory.memsw.limit_in_bytes: No such file or directory
    /tmp/scripts/assemble: line 16: /sys/fs/cgroup/cpu,cpuacct/cpu.cfs_quota_us: No such file or directory

```

A bit of a mystery after some simple searches/greps.

The test errors out at https://github.com/openshift/origin/blob/master/test/extended/builds/s2i_quota.go#L58

This is the test bc:  https://github.com/openshift/origin/blob/master/test/extended/builds/s2i_quota.go#L58

Where that builder image's assemble script is at https://github.com/openshift/build-test-images/blob/master/simples2i/s2i/assemble


Also in the log, we are not finding any memory.max file:

```
Caching blobs under "/var/cache/blobs".
GGM cgroup2 true
GGM file /sys/fs/cgroup/memory.max go err open /sys/fs/cgroup/memory.max: no such file or directory
GGM stat on dir got err <nil>
GGM notExist false
GGM returning default for cgroupv2
I0714 16:32:30.924788       1 builder.go:375] Running build with cgroup limits: api.CGroupLimits{MemoryLimitBytes:0, CPUShares:0, CPUPeriod:0, CPUQuota:0, MemorySwap:0, Parent:""}

```

Even though we've specified limits at https://github.com/openshift/origin/blob/master/test/extended/testdata/builds/test-s2i-build-quota.json#L14-L15

Seems like those should translate to a `memory.max` file and the analogous cpu max file.

Feels like a lower level k8s / crio / cgroup setup error.

WDYT

vrutkovs:
At this point I'm fine with skipping this particular test in cgroupsv2 suite for now.

It appears build quota setting doesn't get passed to cgroups - or builder container can't read it?

adambkaplan:
I assumed that these resource limits get translated into container resource limits. What I am not aware of is how those container resource limits are made visible within the container itself.

gabemontero:
the test works with "cgroups v1", as exhibited by that same test passing in e2e-aws-builds since `/sys/fs/cgroup/memory/memory.limit_in_bytes` is found

What I can't find yet is where that `MEMORY=cat: /sys/fs/cgroup/memory/memory.limit_in_bytes` is coming from.  Seeing where that is would help connect some dots.

It seems like a similar check for cgroups v2 like what we are doing in this PR with util_linux.go is needed where ever that `cat` is.

Perhaps merging this PR to make some progress with the CI tests you are pursuing @vrutkovs is maybe one thing, but
I don't see us handing BUILD-278 off to QE until we see these limits handled properly.

gabemontero:
OK, @nalind @adambkaplan and I had a pow-wow in our team's scrum today.

1) @nalind found the mystery MEMORY=cat: /sys/fs/cgroup/memory/memory.limit_in_bytes source:  https://github.com/openshift/origin/blob/master/test/extended/testdata/builds/build-quota/.s2i/bin/assemble

2) I will need to update that file to echo both the cgroup v1 and v2 locations ... as long as one of the echo's produces the `MEMORY=419430400` string and the MEMORYSWAP one we should  be good

3) that said, there is still a question if the correct cgroupv2 file is getting created / updated when we set resource limits on the build pod.

4) so I'll be posting more debug to traverse in the build pod to traverse `/sys/fs/cgroup` and confirm what its listing looks like compared to @vrutkovs debug at https://github.com/openshift/builder/pull/246#issuecomment-878411890

5) when he frees up from some meeting today @vrutkovs @nalind will be posting some questions to you here specifically that he had around how the nodes are getting set up in your test, to potentially reconcile what we see exactly from the build pod when we traverse `/sys/fs/cggroup`

nalind:
The debug logic @gabemontero added to the builder [here](https://github.com/openshift/builder/pull/252#issuecomment-880090916) was attempting to check the contents of the `/sys/fs/cgroup` directory that it was given, and there didn't appear to be a `memory.max` node there.

On a cgroupv2 node, the container should be seeing the controllers that the runtime configured for it listed in `/sys/fs/cgroup/cgroup.controllers`, along with contents that look more like the container case from https://github.com/openshift/builder/pull/246#issuecomment-878411890, or really any other container.  What do `/sys/fs/cgroup/cgroup.controllers` and `/sys/fs/cgroup` look like in the builder container in a build pod, and if they're different from what we get for a container run using `oc run`, why are they different?

gabemontero:
I've pushed the debug to list the files under `/sys/fs/cgroup` from within the build pod, as well as print the contents of any file that starts with `memory` to confirm whether any of them reflect the memory limit our quota tests places on the build pod.

I'll post data from the debug assuming we get a valid run.

gabemontero:
OK so my latest debug shed some light @vrutkovs @nalind .... see https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_builder/252/pull-ci-openshift-builder-master-e2e-aws-cgroupsv2/1415752738687422464

There are 2 `memory.max` files with the value of `419430400` we were expecting given the quota set on the pod.  Just in kubepod related subdirectories.

GGM /sys/fs/cgroup/kubepods.slice/kubepods-pod37aab7c5_dee2_42cf_833c_f73584920d73.slice/memory.max has contents 419430400
GGM /sys/fs/cgroup/kubepods.slice/kubepods-pod37aab7c5_dee2_42cf_833c_f73584920d73.slice/crio-2e3f3724ae29f2be52003a4ae295a3d472d65e065394ac61a78f2b6675735084.scope/memory.max has contents 419430400


Are we expected to search kubpods.slice for `memory.max` files, or should we expect those files to be linked or copied to `/sys/fs/cgroup/memory.max` ?

Re-reading @nalind 's https://github.com/openshift/builder/pull/252#issuecomment-880890340 I now see I missed the bit about the contents of `/sys/fs/cgroup/cgroup.controllers` and `/sys/fs/cgroup`.  Unless you all already know the answer to my question ^^, I'll assume I need to push another round of debug

pkg/build/builder/util_linux.go
```diff
@@ -16,17 +16,30 @@ func GetCGroupLimits() (*s2iapi.CGroupLimits, error) {
...
+	cgroupV2 := cgroups.IsCgroup2UnifiedMode()
+	if cgroupV2 {
 		file = "/sys/fs/cgroup/memory.max"
 	}
 	byteLimit, err := readMaxStringOrInt64(file)
 
 	if err != nil {
+		_, e := os.Stat("/sys/fs/cgroup")
+		notExist := false
+		if e != nil && os.IsNotExist(e) {
```
kolyshkin, line 34: new code is supposed to use `errors.Is(e, os.ErrNotExist)` but it does not really matter here.

pkg/build/builder/util_linux.go
```diff
@@ -16,17 +16,30 @@ func GetCGroupLimits() (*s2iapi.CGroupLimits, error) {
...
+		}
 		// for systems without cgroups builds should succeed
-		if _, err := os.Stat("/sys/fs/cgroup"); os.IsNotExist(err) {
+		if notExist {
 			return &s2iapi.CGroupLimits{}, nil
 		}
-		return nil, fmt.Errorf("cannot determine cgroup limits: %v", err)
+		// otherwise for cgroupv1 error out
+		if !cgroupV2 {
+			return nil, fmt.Errorf("cannot determine cgroup limits: %v", err)
```
kolyshkin, line 45: Also not important, but maybe use `%w` for error wrapping.

pkg/build/builder/util_linux.go
```diff
@@ -16,17 +16,30 @@ func GetCGroupLimits() (*s2iapi.CGroupLimits, error) {
...
+		}
+		// if the cgroupv2 error is anything other than file does not exists, error out
+		if !notExist {
+			return nil, fmt.Errorf("cannot determine cgroup limits: %v", err)
```
kolyshkin, line 50: ditto

kolyshkin:
If the container is privileged, it most probably have /sys/fs/cgroup bind mounted from the host, and thus sees the same tree as the host, starting from the top cgroup. In cgroup v2, the top cgroup can't have memory limit set.

gabemontero:
On Thu, Jul 15, 2021 at 10:38 PM Kir Kolyshkin ***@***.***>
wrote:

> If the container is privileged, it most probably have /sys/fs/cgroup bind
> mounted from the host, and thus sees the same tree as the host, starting
> from the top cgroup. In cgroup v2, the top cgroup can't have memory limit
> set.
>

I see.  So search for memory.max off of the subdir I mentioned.  The build
container atm is always privileged (though that will change eventually)

Thanks

gabemontero:
> I see. So search for memory.max off of the subdir I mentioned. The build container atm is always privileged (though that will change eventually) Thanks

OK circling back to this today with less multi tasking to distract me @adambkaplan @vrutkovs @nalind @kolyshkin today, my traversing the `/sys/fs/cgroup/kubepods.slice` to find the right `memory.max` file seems untenable.

I don't see where the build container in the build pod has access to the pod uid used in constructing the subdir for the build pod.  Hence, we cannot distinguish between the `memory.max` file for our pod vs. the one for all the other pods on the given host/node.  Every pod on the host is showing up in my debug.  I'm seeing plenty of `memory.max` files with `max` in them.

I suspect that could be one of the items @nalind was skeptical about during our brief build api scrum this AM, but we just did not have time to get into it.

So unless there is something "config wise" that forces things so `/sys/fs/cgroup` is not bind mounted from the host into the build pod when the build pod is privileged, then I think we are stuck wrt ccgroup v2.  Or unless there is some means of finding the "right" `memory.max` file for our privileged pod that has not been discuss amongst us here.

If so, I think that at most leaves us with
1) @vrutkovs skipping the s2i quota test per his comment in https://github.com/openshift/builder/pull/252#issuecomment-880109754
2) we pull the Jira BUILD-279 from the sprint/release @adambkaplan until either we can get things sorted out for privileged pods/containers, or we get unprivileged builds in, and then cite that we only support unprivileged builds for cgroupsv2

In addition to any discourse here, I'll bring this up during our team arch meeting tomorrow or team office hours on Wednesday (I'm posting this on a Monday)

gabemontero:
I've pushed an update that clears out the temp dbg, includes the fixes to get to the point where only the quota test fails, and includes updates from @kolyshkin last set of comments.

We still have to make the decision wrt how cgroup2 works with privileged pods, do we stop short of pulling full BUILD-278 completion for the time being, that I mentioned in https://github.com/openshift/builder/pull/252#issuecomment-882810384

nalind:
> I suspect that could be one of the items @nalind was skeptical about during our brief build api scrum this AM, but we just did not have time to get into it.

I did a little bit more research into what `/sys/fs/cgroup` looks like:
* for an unprivileged container, it's the cgroup of the container itself, so we want to look at the `/sys/fs/cgroup/memory.max`
* for a privileged container, it's the `/sys/fs/cgroup` from the node, so if we're aware of that, we need to scan `/proc/self/cgroup` for the name of the subdirectory under `/sys/fs/cgroup` in which the `memory.max` that applies to us can be found.

gabemontero:
> > I suspect that could be one of the items @nalind was skeptical about during our brief build api scrum this AM, but we just did not have time to get into it.
> 
> I did a little bit more research into what `/sys/fs/cgroup` looks like:
> 
>     * for an unprivileged container, it's the cgroup of the container itself, so we want to look at the `/sys/fs/cgroup/memory.max`
> 
>     * for a privileged container, it's the `/sys/fs/cgroup` from the node, so if we're aware of that, we need to scan `/proc/self/cgroup` for the name of the subdirectory under `/sys/fs/cgroup` in which the `memory.max` that applies to us can be found.

Ahh ... that definitely feels tenable @nalind 

Thanks for the info / research!  I'll then circle back to this, add some debug to research exactly what file under `/proc/self/cgroup` and see if the dots in fact can be connected.

gabemontero:
In particular I see https://www.kernel.org/doc/html/latest/admin-guide/cgroup-v2.html#basics @nalind 

I'll cat the contents of the file, and see if the "id" there corresponds to any subdirs I see under /sys/fs/cgroup

nalind:
I'd also recommend the NOTES section of the cgroups(7) manual page for a quick rundown of what we should expect to find in `/proc/self/cgroup`.  The map that `github.com/opencontainers/runc/libcontainer/cgroups.ParseCgroupFile("/proc/self/cgroup")` returns should have a path for `""` in it when `IsCgroup2UnifiedMode()` is `true`.

gabemontero:
ok https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_builder/252/pull-ci-openshift-builder-master-e2e-aws-cgroupsv2/1417477246209232896 just has the quota failure 

so no regressions since I removed my last round of debug and addressed @kolyshkin 's comments

working off of @nalind 's suggestions

gabemontero:
AWS fail on cgroups2

/test e2e-aws-cgroupsv2

gabemontero:
OK, next layer of the onion.

The code is now finding the correct memory.max file with the expected value:

`2021-07-21T16:39:50.976759577Z I0721 16:39:50.976718       1 builder.go:335] using /sys/fs/cgroup/kubepods.slice/kubepods-pod2d3934aa_cd06_48df_a5ca_fe0f2ca32cdc.slice/crio-8325ee4c949390b64c0f7d887db96c538ba9c2f6071991138189daf04248c3cf.scope/memory.max for cgroup2 memory.max with value 419430400`

But subsequent processing for the parent fails with: 

`2021-07-21T16:39:51.378746286Z F0721 16:39:51.378713       1 helpers.go:115] error: failed to retrieve cgroup limits: read cgroup parent: could not find memory cgroup subsystem in map map[:/kubepods.slice/kubepods-pod2d3934aa_cd06_48df_a5ca_fe0f2ca32cdc.slice/crio-8325ee4c949390b64c0f7d887db96c538ba9c2f6071991138189daf04248c3cf.scope]`

Going back to the various doc/man pages to see what is the cgroup v2 way to determine parents.

gabemontero:
Yeah with "default" cgroup v1 there is a lot more entries in `/proc/self/cgroup`.  There is only the 1 entry in the cgroup v2 test.

v1:


`I0721 18:02:46.705323       1 util_linux.go:38] found cgroup values map: map[blkio:/kubepods.slice/kubepods-pod70aeabdb_f767_4c2d_b2bc_304e0067e7b5.slice/crio-a45b844fd3f8a28ca0719590ece152c99436346144d39a8b1268bc9f6314b094.scope cpu:/kubepods.slice/kubepods-pod70aeabdb_f767_4c2d_b2bc_304e0067e7b5.slice/crio-a45b844fd3f8a28ca0719590ece152c99436346144d39a8b1268bc9f6314b094.scope cpuacct:/kubepods.slice/kubepods-pod70aeabdb_f767_4c2d_b2bc_304e0067e7b5.slice/crio-a45b844fd3f8a28ca0719590ece152c99436346144d39a8b1268bc9f6314b094.scope cpuset:/kubepods.slice/kubepods-pod70aeabdb_f767_4c2d_b2bc_304e0067e7b5.slice/crio-a45b844fd3f8a28ca0719590ece152c99436346144d39a8b1268bc9f6314b094.scope devices:/kubepods.slice/kubepods-pod70aeabdb_f767_4c2d_b2bc_304e0067e7b5.slice/crio-a45b844fd3f8a28ca0719590ece152c99436346144d39a8b1268bc9f6314b094.scope freezer:/kubepods.slice/kubepods-pod70aeabdb_f767_4c2d_b2bc_304e0067e7b5.slice/crio-a45b844fd3f8a28ca0719590ece152c99436346144d39a8b1268bc9f6314b094.scope hugetlb:/kubepods.slice/kubepods-pod70aeabdb_f767_4c2d_b2bc_304e0067e7b5.slice/crio-a45b844fd3f8a28ca0719590ece152c99436346144d39a8b1268bc9f6314b094.scope memory:/kubepods.slice/kubepods-pod70aeabdb_f767_4c2d_b2bc_304e0067e7b5.slice/crio-a45b844fd3f8a28ca0719590ece152c99436346144d39a8b1268bc9f6314b094.scope name=systemd:/kubepods.slice/kubepods-pod70aeabdb_f767_4c2d_b2bc_304e0067e7b5.slice/crio-a45b844fd3f8a28ca0719590ece152c99436346144d39a8b1268bc9f6314b094.scope net_cls:/kubepods.slice/kubepods-pod70aeabdb_f767_4c2d_b2bc_304e0067e7b5.slice/crio-a45b844fd3f8a28ca0719590ece152c99436346144d39a8b1268bc9f6314b094.scope net_prio:/kubepods.slice/kubepods-pod70aeabdb_f767_4c2d_b2bc_304e0067e7b5.slice/crio-a45b844fd3f8a28ca0719590ece152c99436346144d39a8b1268bc9f6314b094.scope perf_event:/kubepods.slice/kubepods-pod70aeabdb_f767_4c2d_b2bc_304e0067e7b5.slice/crio-a45b844fd3f8a28ca0719590ece152c99436346144d39a8b1268bc9f6314b094.scope pids:/kubepods.slice/kubepods-pod70aeabdb_f767_4c2d_b2bc_304e0067e7b5.slice/crio-a45b844fd3f8a28ca0719590ece152c99436346144d39a8b1268bc9f6314b094.scope rdma:/]`
`I0721 18:02:46.705381       1 util_linux.go:64] cgroup memory subsystem value: /kubepods.slice/kubepods-pod70aeabdb_f767_4c2d_b2bc_304e0067e7b5.slice/crio-a45b844fd3f8a28ca0719590ece152c99436346144d39a8b1268bc9f6314b094.scope`
`I0721 18:02:46.705409       1 util_linux.go:64] found cgroup parent kubepods-pod70aeabdb_f767_4c2d_b2bc_304e0067e7b5.slice`

v2: 

`2021-07-21T16:39:51.372202562Z I0721 16:39:51.372160       1 util_linux.go:103] found cgroup values map: map[:/kubepods.slice/kubepods-pod2d3934aa_cd06_48df_a5ca_fe0f2ca32cdc.slice/crio-8325ee4c949390b64c0f7d887db96c538ba9c2f6071991138189daf04248c3cf.scope]`


The parent information is ultimately pulled from the kv pair when cgroupv1:

`memory:/kubepods.slice/kubepods-pod70aeabdb_f767_4c2d_b2bc_304e0067e7b5.slice/crio-a45b844fd3f8a28ca0719590ece152c99436346144d39a8b1268bc9f6314b094.scope name=systemd:/kubepods.slice/kubepods-pod70aeabdb_f767_4c2d_b2bc_304e0067e7b5.slice/crio-a45b844fd3f8a28ca0719590ece152c99436346144d39a8b1268bc9f6314b094.scope`

gabemontero:
/hold cancel

today between @nalind @adambkaplan and myself we figured out we don't need any of the parent group ref stuff 

it is a holdover from 3.x when the docker daemon was in play

otherwise, once the redhat.com outage is sorted out, other than merging some form of https://github.com/openshift/origin/pull/26338 to accommodate both cgroup versions both e2e-aws-builds and e2e-aws-cgroupsv1 should pass here

the PR is ready for final review induced coding tweaks and possible tags for merge

gabemontero:
internet outage is suppose to be fixed according to RH wide announcement

/retest

gabemontero:
@rolfedh @jitendar-singh @xiuwang @wewang58 @sferich888 - as we are past feature freeze, I believe doc / qe / px labels from you all will count as the "valid bz" req for this

If I read the latest no feature freeze updates correctly

@rolfedh per my Jira update IMO BUILD-278 should be nodoc

@adambkaplan ^^

adambkaplan:
/label docs-approved

No docs are required

adambkaplan:
/label px-approved

No customer enablement is required for this feature

adambkaplan:
/refresh

This repo shouldn't require the valid-bug label

adambkaplan:
So I double checked prow - it is indeed configured correctly. The tide message that appears on GitHub is misleading.

gabemontero:
OK so the debug from the latest openshift/origin quota tests was informative.

We have `memory.max` in these subfolders underneath where the openshift/origin test is currently looking:

```
/sys/fs/cgroup/user.slice/memory.max
/sys/fs/cgroup/pids/memory.max
/sys/fs/cgroup/init.scope/memory.max
```

followed by a bunch under `/sys/fs/cgroup/system.slice` where there are various systemd subcomponents and services listed in subdirectories; going out on a limb and thinking those are not relevant.

I'm guessing those are there though because of the previous revelation from @kolyshkin in that since our pod/container is privileged, we get a bunch of stuff from the host/node.

Similarly, there are a bunch of instances under `/sys/fs/cgroup/kubepods.slice/kubepods-burstable.slice/kubepods-burstable-pod*` for the various pods on the node.  But my sense is our openshift/origin test should not have to deal with pod mapping like we have to do in the builder image here with this PR.  So I am dismissing those possibilities as well, except for perhaps the top level

`/sys/fs/cgroup/kubepods.slice/memory.max` or `/sys/fs/cgroup/kubepods.slice/kubepods-besteffort.slice/memory.max`

But with everything I listed, it seems like either `/sys/fs/cgroup/user.slice/memory.max` or `/sys/fs/cgroup/pids/memory.max` is the one.

I did go back to the reference https://www.kernel.org/doc/html/latest/admin-guide/cgroup-v2.html#basics that @nalind pointed me to and could not however find references to either of those two locations for `memory.max`.

Any thoughts @nalind or @kolyshkin ?

In the interim, I am going to add some temporary debug back into this PR to see if either of those files are present and visible to the us we are are processing in util_linux.go, and see if they have the expected value.  If we get a match, I'll update the openshift/origin test with the final solution.  Otherwise, I'm going to need to merge another temporary change to the openshift origin tests where I echo the name of the file, then cat the contents, so we can nail down which `memory.max` should be inspected.

gabemontero:
> But with everything I listed, it seems like either `/sys/fs/cgroup/user.slice/memory.max` or `/sys/fs/cgroup/pids/memory.max` is the one.
> 
> I did go back to the reference https://www.kernel.org/doc/html/latest/admin-guide/cgroup-v2.html#basics that @nalind pointed me to and could not however find references to either of those two locations for `memory.max`.
> 
> Any thoughts @nalind or @kolyshkin ?
> 
> In the interim, I am going to add some temporary debug back into this PR to see if either of those files are present and visible to the us we are are processing in util_linux.go, and see if they have the expected value. If we get a match, I'll update the openshift/origin test with the final solution. Otherwise, I'm going to need to merge another temporary change to the openshift origin tests where I echo the name of the file, then cat the contents, so we can nail down which `memory.max` should be inspected.

Unfortunately, only the pod specific memory max files had the correct value.  All those top level ones either had `max` or much larger numbers which looked like the max amount of mem on the node.

So the bash script test on the openshift/origin side will to search for the `memory.max` file that has the correct amount under kubepods.slice

like `/sys/fs/cgroup/kubepods.slice/kubepods-pod1460b72e_88ad_4d68_b63a_e579fa7daae9.slice/memory.max`

sigh

gabemontero:
So the latest test change from https://github.com/openshift/origin/pull/26395 was able to post the expected setting of `MEMORY` when quota was applied.

However, `MEMORYSWAP` did not show up.  And with the debug present so far, I'm not seeing an analogous setting of swap for cgroupv2 like we did with cgroupv1 in the memory swap files found under /sys/fs/cgroup.

Will need to circle back with the doc previously cited and the folks on this PR to see if that is expected for cgroupv2, or if we need to investigate further.

gabemontero:
> So the latest test change from [openshift/origin#26395](https://github.com/openshift/origin/pull/26395) was able to post the expected setting of `MEMORY` when quota was applied.
> 
> However, `MEMORYSWAP` did not show up. And with the debug present so far, I'm not seeing an analogous setting of swap for cgroupv2 like we did with cgroupv1 in the memory swap files found under /sys/fs/cgroup.
> 
> Will need to circle back with the doc previously cited and the folks on this PR to see if that is expected for cgroupv2, or if we need to investigate further.

OK after some research on my end and an exchange with @nalind on slack, confirmed that the expectaions for swap in v1 is not the same as v2.  With v1 `memory.limit_in_bytes)` is memory+swap, with v2 `memory.swap.max` is just swap.  So for v2 we should see a `0`

And I do see that with my latest debug.

So we need one more openshift/origin PR and a tweak to the quota assemble script to account for this difference.

gabemontero:
cgroupv2 passed !!!

and the e2e-aws-builds was a sig-etcd flake ... all the sig-builds passed !!!

going to remove the debug commit and see about getting clean e2e's next go around and getting lgtm / approve labels from the reviewers :-)

gabemontero:
sig-builds passed

unrelated flakes in cgroupv2

/test e2e-aws-cgroupv2

adambkaplan (approved):
/lgtm

gabemontero:
/assign @jitendar-singh 

Hey @jitendar-singh  - so we are ready to talk about what level of verification you may or may not need to do with this, and thus how comfortable are you with applying the qe-approved label.

At its core, this change just adds the memory limit from the k8s pod mem resource limit to the appropriate option field for buildah before we call it.

buildah ultimately passes that to the opencontainers code, which basically sets the linux level mem limits.

So,

1) how to set up with cgroupv2 ... is it viable for you to install a cluster in that fashion?

2) what is needed / possible to satisfy that buildah/opencontainers/linux has the memory limit?

For 1), the best I can tell you is that you have to install a cluster and then apply the necessary machine config operator changes to have it boot in cgroup v2 mode, like the e2-aws-ccgroupsv2 job that @vrutkovs set up for us does ... running this PRs changes against that job is how I did development here.

@vrutkovs please assist if I'm missing key details there.

What I found around that machine config change and setting the required kernel arguments is https://github.com/openshift/release/blob/master/ci-operator/step-registry/openshift/manifests/cgroupsv2/openshift-manifests-cgroupsv2-commands.sh

get those machine config yaml saved, then per the OCP doc apply those machine config chagnes and reboot the nodes/cluster as needed for them to take effect, with the underlying hosts now running in cgroupv2 mode.

For 2), unless running with maxium trace on a build prints sufficient data to confirm the memory limit is set, I think we need to link up with @nalind and see if the inspection of the /sys/fs/cgroup like in https://github.com/openshift/origin/blob/master/test/extended/testdata/builds/build-quota/.s2i/bin/assemble is sufficient, or if there are some additional verification in the build container we should do.

Or for both 1) and 2), we say that since cgroupv2 is a dev/tech preview only, the PR verification is sufficient for QE coverage on this?

Hopefully we can discuss during scrum / office hours on Monday/Tuesday next week.

thanks

gabemontero:
quick update:  based on the discussion (for RH internal only) in https://coreos.slack.com/archives/C02258G4S79/p1629482111160000 we seem to be leaning toward PR validation being sufficient and just applying the QE label and merging, but I'm still awaiting on whether we have consensus with @jitendar-singh and @adambkaplan

adambkaplan:
/label qe-approved

Since cgroupsv2 isn't even tech preview yet, CI tests passing is sufficient here.

PR #254, opened 2021-07-14
unprivileged: don't try to set up user namespaces for manage-dockerfile/git-clone

nalind:
Walk back a bit of #220: don't assume that we can set up a user namespace when we're invoked as openshift-manage-dockerfile or openshift-git-clone, even when not explicitly given mappings to use, since those containers haven't traditionally been run with privileged=true.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -77,12 +77,17 @@ func main() {
 	flags.BoolVar(&useNewgidmap, "use-newgidmap", os.Geteuid() != 0, "use newgidmap to set up GID mappings")
 	wrapped := command.Run
 	command.Run = func(c *cobra.Command, args []string) {
-		storeOptions, err := storage.DefaultStoreOptions(false, 0)
-		kcmdutil.CheckErr(err)
-		os.MkdirAll(storeOptions.GraphRoot, 0775)
-		os.MkdirAll(storeOptions.RunRoot, 0775)
-		maybeReexecUsingUserNamespace(uidmap, useNewuidmap, gidmap, useNewgidmap)
-		wrapped(c, args)
+		switch basename {
+		case "openshift-sti-build", "openshift-docker-build", "openshift-extract-image-content":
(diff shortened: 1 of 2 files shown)


gabemontero:
@vrutkovs FYI ... eventually, I wonder if we should create an optional OKD version of e2e-aws-builds in this repo to catch whatever problem you saw in the nightlies .... WDYT?

nalind:
I'm unlikely to ever argue against adding tests for a thing.  How, and how far, does the OKD version diverge from the existing e2e-aws-builds test, though?

gabemontero:
On the surface @nalind 

/lgtm

now we need @adambkaplan for the approve label.

And, per the feedback from @vrutkovs I'm wondering if there is anything besides adding the equivalent of https://github.com/openshift/release/blob/master/ci-operator/jobs/openshift/cluster-samples-operator/openshift-cluster-samples-operator-master-presubmits.yaml#L505-L573 for openshift/builder so we can get some OKD validation for this.

An optional okd-e2e-aws-builds that we could launch from this PR.

gabemontero:
> I'm unlikely to ever argue against adding tests for a thing. How, and how far, does the OKD version diverge from the existing e2e-aws-builds test, though?

I'll defer to @vrutkovs ... but presumably enough for the OKD nightlies to flag something.

vrutkovs:
>How, and how far, does the OKD version diverge from the existing e2e-aws-builds test, though?

The main difference is FCOS is used in `machine-os-content`, so host system is Fedora.
Other changes are insignificant (community operators only, custom installer using FCOS and such).

I can create okd-specific builder test, but we're hitting a DPTP bug in 4.9 rehearsals

adambkaplan (approved):
/approve

vrutkovs:
That worked - https://amd64.origin.releases.ci.openshift.org/releasestream/4.9.0-0.okd/release/4.9.0-0.okd-2021-07-14-215804, thank you!

PR #255, opened 2021-07-14
Let the serviceability package determine how to tie logrus to klog

nalind:
Let the serviceability package decide how to map klog logging levels to logrus levels.  Per [InitLogrusFromKlog](https://github.com/openshift/library-go/blob/9396bd4bf9966d7c7827d8eeb38d80b6bb311cb6/pkg/serviceability/logrus.go#L13), this has the effect of changing the logging level assumed for debug-level messages from 6 to 4.


adambkaplan (approved):
/lgtm

This will give us a fuller picture when we ask customers to run builds with log level = 5.

PR #256, opened 2021-07-16
Dockerfiles: run `rpm --setcaps shadow-utils` during build

nalind:
Our base images don't preserve file capabilities on `/usr/bin/newuidmap` and `/usr/bin/newgidmap`, but they do preserve setuid/setgid bits, which grant more privileges to callers, so go ahead and restore file capabilities during the build.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,6 +11,7 @@ RUN INSTALL_PKGS=" \
       " && \
     yum install -y --setopt=skip_missing_names_on_install=False $INSTALL_PKGS && \
     yum clean all
+RUN rpm --setcaps shadow-utils
 COPY --from=builder /go/src/github.com/openshift/builder/openshift-builder /usr/bin/
 COPY imagecontent/bin /usr/bin
 COPY imagecontent/etc/containers /etc/containers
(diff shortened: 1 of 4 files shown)


rhatdan:
What exactly does rpm --setcaps do to shadow-utils?  Does it add the filecaps back in?

rhatdan:
Ubuntu man pages says:
```
       rpm --setcaps PACKAGE_NAME
              sets capabilities of files in the given package. Consider using --restore instead.

       rpm --restore PACKAGE_NAME
              The option restores owner, group, permissions and  capabilities  of  files  in  the
              given package.
```

nalind:
> What exactly does rpm --setcaps do to shadow-utils? Does it add the filecaps back in?

Yes, it resets the file capabilities on files from that package, particularly `/usr/bin/newuidmap` and `/usr/bin/newgidmap`.  The base image (so far as I can tell) didn't preserve the `security.capability` attributes of its contents.  The ownerships and DAC permissions appear to be preserved, including the `setuid` bit on binaries which should have them.

adambkaplan:
@nalind I believe we are still working on getting builds to pass with cgroupsv2. Do we explicitly need that to test this update?

nalind:
No, whether or not the `newuidmap` and `newgidmap` binaries in the builder image have their `security.capability` xattr set correctly shouldn't affect the cgroupsv2 test's result, since that test is still going to be running the build container with privileges, and they're not invoked in that case.  This is a guess at part of what's keeping #173 from passing.

nalind:
Perhaps we can merge this, and revert it if it starts causing errors at image build time?

coreydaley:
Looks like it's just waiting on the qe-approved label.

PR #266, opened 2021-10-14
Bug 2011293: add Red Hat registries to the default registries.conf included in the builder image

gabemontero:


gabemontero:
> Syntax looks fine from here, though we may want to take the opportunity to replace this version, which is in version 1 of the registries.conf file format, with one that's in version 2 of the format. (The code base can accept either version.) Examples:
> 
>     * https://src.fedoraproject.org/rpms/containers-common/raw/rawhide/f/registries.conf (Fedora Raw Hide)
> 
>     * https://gitlab.com/redhat/centos-stream/rpms/containers-common/-/raw/stream-container-tools-latest-rhel-9.0.0-beta/registries.conf (CentOS Stream 9)
> 
>     * https://raw.githubusercontent.com/containers/image/main/registries.conf (containers/image main branch)

seems reasonable to me @nalind and thanks for those refs

I'm first going to launch a cluster with cluster-bot including this PR, just to sanity check our existing change here with my reproducer

Assuming all is well, I'll then devise as a separate commit a v2 registries.conf, then do the same.  If not, debug.

If v2 is still well, squash commits, and we tag for merge

If not, remove v2 commit and we tag for merge, unless there is some obvious / minor tweak that is needed

nalind:
Is `docker.io/library/rhel7.7` a real image?  Even when logged in to docker hub, I can't `skopeo inspect` it.

gabemontero:
OK this clusterbot launch worked and was able to verify that registry search was fully employed when we removed the docker:// prefix.

I'll give the conversion to v2 registries.conf a go on Monday and we'll go from there.

gabemontero:
> LGTM. The default for Fedora's containers-common package, once the commented lines are removed, looks like:
> 
> ```
> unqualified-search-registries = ["registry.fedoraproject.org", "registry.access.redhat.com", "docker.io", "quay.io"]
> short-name-mode="enforcing"
> ```
> 
> I gather we want to use the default `short-name-mode` of `permissive`, so omitting the setting like the PR currently does is fine, and the list of registries in the PR may be a better set of defaults for builds in OpenShift.

thanks @nalind

per last week I've also launched a clusterbot cluster with this PR, and will try the reproducing test case manually if that cluster comes up successfully

imagecontent/etc/containers/registries.conf
```diff
@@ -1,30 +1,22 @@
...
+# myregistry.com is not first in the search list, an attacker could place a
+# different `foobar` image at a registry earlier in the search list. The user
+# would accidentally pull and run the attacker's image and code rather than the
+# intended content. We recommend only adding registries which are completely
+# trusted (i.e., registries which don't allow unknown or anonymous users to
+# create accounts with arbitrary names). This will prevent an image from being
+# spoofed, squatted or otherwise made insecure.  If it is necessary to use one
+# of these registries, it should be added at the end of the list.
+#
+# # An array of host[:port] registries to try when pulling an unqualified image, in order.
```
adambkaplan, line 20: I highly recommend folks watch the "Registries after Dark" talk from KubeCon NA 2021 once the video is published. Name squatting is a very real threat.

adambkaplan (commented):
I wonder if the build controller is still providing us with a v1 formatted `registries.conf` file. There is a separate feature in flight to provide the v2 formatted file, since that will let us mirror images by tag.

gabemontero:
> > > > I gather we want to use the default `short-name-mode` of `permissive`, so omitting the setting like the PR currently does is fine, and the list of registries in the PR may be a better set of defaults for builds in OpenShift.
> 
> Looks like that requires `short-name-mode="disabled"` after all.

ok I'll add `short-name-mode="disabled"` and clusterbot the PR one more time and see if that works 

that said, per @adambkaplan 's https://github.com/openshift/builder/pull/266#pullrequestreview-782503434 with a separate feature in flight to do v2 registries.conf it probably makes sense to not go to v2 in the bz pr anyway.

but I'm curious, so will try the last thing @nalind

gabemontero:
> > > > > I gather we want to use the default `short-name-mode` of `permissive`, so omitting the setting like the PR currently does is fine, and the list of registries in the PR may be a better set of defaults for builds in OpenShift.
> > 
> > 
> > Looks like that requires `short-name-mode="disabled"` after all.
> 
> ok I'll add `short-name-mode="disabled"` and clusterbot the PR one more time and see if that works
> 
> that said, per @adambkaplan 's [#266 (review)](https://github.com/openshift/builder/pull/266#pullrequestreview-782503434) with a separate feature in flight to do v2 registries.conf it probably makes sense to not go to v2 in the bz pr anyway.
> 
> but I'm curious, so will try the last thing @nalind

ok short-name-mode="disabled" worked @nalind  - thanks 

so assuming the e2e's are OK, @adambkaplan WDYT about switching to v2 with this PR? ... @coreydaley fyi since you approved this

imagecontent/etc/containers/registries.conf
```diff
@@ -1,30 +1,23 @@
...
+unqualified-search-registries = ["registry.redhat.io","registry.access.redhat.com", "quay.io", "docker.io"]
```
coreydaley, line 21: Do we need any type of test to verify the search order of the registries?

imagecontent/etc/containers/registries.conf
```diff
@@ -1,30 +1,23 @@
...
-# Blocked Registries, blocks the `docker daemon` from pulling from the blocked registry.  If you specify
-# "*", then the docker daemon will only be allowed to pull from registries listed above in the search
-# registries.  Blocked Registries is deprecated because other container runtimes and tools will not use it.
-# It is recommended that you use the trust policy file /etc/containers/policy.json to control which
-# registries you want to allow users to pull and push from.  policy.json gives greater flexibility, and
-# supports all container runtimes and tools including the docker daemon, cri-o, buildah ...
-# The atomic CLI `atomic trust` can be used to easily configure the policy.json file.
-[registries.block]
-registries = []
+short-name-mode="disabled"
```
coreydaley, line 23: Or a test to make sure that this does not get removed accidentally and nobody notices?

coreydaley: Sounds good to me, you can close this item.

gabemontero: https://github.com/openshift/origin/pull/26530 is up

gabemontero:
> /hold
> 
> I think there are security implications to this change. Two years ago (eep!) I submitted a PR to try and implement a similar change [1]. This was ultimately closed because the way buildah worked at the time, it would potentially send a `docker.io` pull secret to another container registry.
> 
> With time plenty has changed in buildah's codebase. Are we able to ensure that as buildah tries to pull from the registries in the search list, it will only use credentials that are appropriate for each individual registry?
> 
> [1] [#40 (comment)](https://github.com/openshift/builder/pull/40#issuecomment-459159014)

WDYT @nalind ^^ ?

gabemontero:
Also, if we don't fix this somehow, we *definitely* need to update the docs to explain what configuration the cluster admin will need to do to accommodate this scenario.  Perhaps a KCS article as well.

@rolfedh FYI

rolfedh:
> Also, if we don't fix this somehow, we _definitely_ need to update the docs to explain what configuration the cluster admin will need to do to accommodate this scenario. Perhaps a KCS article as well.
> 
> @rolfedh FYI

This sounds like a known issue in versions 4.6+ with a workaround, right?, Please use [this document](https://docs.google.com/document/d/1dvxvWteHi4cBbhpL9yeHd6x8OXVQPPCEO_uscbeT_qU/edit?usp=sharing) to describe the workaround and then ping me.

PR #270, opened 2021-11-09
Bug 2021551: getAssembleUser(): strip the group part out before checking the UID

nalind:
If the s2i assemble image's configured user includes a group (i.e., the user spec is in the form "user:group"), strip the group part before checking if the UID is in the range of allowed users.  Borrow `extractUsers()` from `github.com/openshift/source-to-image/pkg/docker` to handle it.

### DIFF
--- a/pkg/build/builder/sti.go
+++ b/pkg/build/builder/sti.go
@@ -600,5 +600,13 @@ func getAssembleUser(docker DockerClient, imageTag string) (string, error) {
 		// If the builder image has the assemble-user label, override with the provided value
 		assembleUser = labelAssembleUser
 	}
-	return assembleUser, nil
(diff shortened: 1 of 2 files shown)


gabemontero:
at least to answer my question in slack about this, yeah, the 3.11 version of this path at https://github.com/openshift/origin/blob/release-3.11/pkg/build/builder/sti.go does not employ `getAssembleUser` or the `AssembleUser` field in the s2i config object.

that at least helps explain the apples to oranges comparison and why it could have worked in 3.x but not 4.x

gabemontero:
/lgtm

/assign @adambkaplan 

any objections @adambkaplan ?

adambkaplan (approved):
/lgtm

PR #274, opened 2021-11-17
Bug 2026589: [release-4.9] Update containers/storage to address incorrect overlay options being set on rhel7 nodes

nalind:
Bump github.com/containers/storage to have it correctly diagnose when globally-set options will cause problems for the overlay driver.
Pin github.com/opencontainers/runc to v1.0.0-rc95, as the current version moved API that our pinned version of github.com/docker/docker depended on.


gabemontero:
the RHEL SCL team introduce a regression in their python image for hot deploy @nalind that is leading to the one failure in image-ecosystem ... your changes are not causing that 

so let's if we  can

/skip

since image eco is not required but optional

nalind:
/retitle Bug 2026589: Update containers/storage to address incorrect overlay options being set on rhel7 nodes

nalind:
/retitle Bug 2026589: [release-4.9] Update containers/storage to address incorrect overlay options being set on rhel7 nodes

otaviof:
/test e2e-aws-builds-rhel7

Trying again due timeout on pulling `registry.access.redhat.com/ubi8/ruby-27:latest`, which I can do from my workstation normally, for the record.

adambkaplan:
/close

There is a delta between this PR and the cherrypick in #279 - going to use the latter.

PR #291, opened 2022-02-23
BUILD-433: use the right mappings when we don't need to set any

nalind:
When we create a new user namespace so that we can get CAP_SYS_ADMIN over a mount namespace, if we're already UID 0, map the current set of ID mappings over themselves instead of trying to repeat the mapping that's been applied to the container we're in.

If we know we've been started in a namespace that doesn't have node-default mappings, log that we know.

Don't default to trying "metacopy=on" with "mount_program=fuse-overlayfs", since it knows nothing about the option and all we do is trigger a repeated warning about it not being recognized.

### DIFF
--- a/cmd/userns.go
+++ b/cmd/userns.go
@@ -64,6 +64,13 @@ func parseIDMappings(uidmap, gidmap string) ([]specs.LinuxIDMapping, []specs.Lin
 		if err != nil {
 			klog.Fatalf("Error reading ID mappings for %s: %v\n", uid, err)
 		}
+	} else {
+		for i := range UIDs {
+			UIDs[i].HostID = UIDs[i].ContainerID
+		}
+		for i := range GIDs {
+			GIDs[i].HostID = GIDs[i].ContainerID
+		}
 	}
 	if uidMappings := parseMapping("uidmap", uidmap); len(uidMappings) != 0 {
 		UIDs = uidMappings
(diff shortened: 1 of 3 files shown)


gabemontero:
/assign @coreydaley 

he has approval for builder repo now

cmd/userns.go
```diff
@@ -91,6 +98,33 @@ func maybeReexecUsingUserNamespace(uidmap string, useNewuidmap bool, gidmap stri
...
 	}
 
+	// Log the ID maps we were started with.
+	UIDs, GIDs, err := unshare.GetHostIDMappings("")
+	if err != nil {
+		klog.Fatalf("Error reading current ID mappings: %v\n", err)
+	}
+	isNodeDefaultMapping := func(m []specs.LinuxIDMapping) bool {
+		return len(m) == 1 && m[0].ContainerID == 0 && m[0].HostID == 0 && m[0].Size == 0xffffffff
+	}
```
adambkaplan, line 108: Style nit - I prefer we move this to a regular function so that we can document with a unit test.

cmd/userns.go
```diff
@@ -91,6 +98,33 @@ func maybeReexecUsingUserNamespace(uidmap string, useNewuidmap bool, gidmap stri
...
+		}
+		for i, entry := range GIDs {
+			if i > 0 {
+				gidMap += ", "
+			}
+			gidMap += fmt.Sprintf("(%d:%d:%d)", entry.ContainerID, entry.HostID, entry.Size)
+		}
+		uidMap += "]"
+		gidMap += "]"
+		klog.Infof("Started in kernel user namespace with UID map %s and GID map %s.", uidMap, gidMap)
```
adambkaplan, line 125: Is this something that we always want to log, log at debug level (5), or somewhere in between (2)?

nalind: I don't personally have a preference.  It's mainly there so that we can test for it, but the test can also adjust the logging level.  Do you have a preference?

cmd/userns.go
nalind: Pulled it out as a separate function.

adambkaplan:
/assign @rolfedh 

For docs approval - I drafted docs in https://github.com/openshift/openshift-docs/pull/44000

adambkaplan (approved):
/lgtm

adambkaplan:
/cc @jitendar-singh @prietyc123 

For QE approval

adambkaplan:
/cc @RickJWagner 

For PX approval

adambkaplan:
/label docs-approved

This PR doesn't impact builds as is, and is not activated until https://github.com/openshift/openshift-controller-manager/pull/173 merges. Furthermore, there won't be official docs for this feature because we are releasing it as Developer Preview.

PR #294, opened 2022-04-05
Bug 2067775: Update prometheus client_golang from 1.11.0 => 1.11.1

coreydaley:


go.mod
```diff
@@ -34,6 +34,8 @@ require (
 	k8s.io/kubernetes v1.23.0
 )
 
+require github.com/stretchr/testify v1.7.0
```
gabemontero, line 37: same as your other PR .... either put this with the other requires, or put a comment explaining why it is separated

coreydaley: updated.

go.mod
```diff
@@ -193,6 +193,7 @@ replace (
 	github.com/docker/docker => github.com/docker/docker v0.0.0-20200911110540-7ca355652fe0
 	github.com/docker/libnetwork => github.com/docker/libnetwork v0.0.0-20210525090646-64b7a4574d14
 	github.com/godbus/dbus => github.com/godbus/dbus v0.0.0-20181101234600-2ff6f7ffd60f
+	github.com/prometheus/client_golang => github.com/prometheus/client_golang v1.11.1 // CVE-2022-21698
```
adambkaplan, line 196: Do we know where we are pulling this as an indirect dependency? For instance, if we are getting it from k8s, we should see if there is a bugfix to the k8s libraries.

coreydaley: I looked at the k8s libraries and the latest version was not updated yet.

adambkaplan: Running `go mod graph | grep "prometheus/client_golang"`, lots of things depend on client_golang.
One of them is openshift/library-go - we could do the following:

1. Update library-go with the dependency update.
2. Re-vendor library-go here.

https://github.com/openshift/library-go/blob/master/go.mod#L24

coreydaley: Opened https://github.com/openshift/library-go/pull/1349 and will update this pull request once that merges

coreydaley:
/hold
Waiting on https://github.com/openshift/library-go/pull/1349

coreydaley:
@adambkaplan Updated library-go , ptal

adambkaplan (approved):
/lgtm

PR #295, opened 2022-04-05
Bug 2069512: Update imgcrypt 1.1.1 => 1.1.4

coreydaley:


go.mod
```diff
@@ -34,6 +34,8 @@ require (
 	k8s.io/kubernetes v1.23.0
 )
 
+require github.com/stretchr/testify v1.7.0
```
gabemontero, line 37: assuming this was separated out for a reason @coreydaley ?

if so, can you put a comment here as to why?

coreydaley: Weird, go mod must have done that. I have merged it into the larger require section and pushed the commit.
Good catch!

go.mod
```diff
@@ -190,6 +190,7 @@ require (
 replace (
 	// temporary replacement for development purposes
 	github.com/containerd/containerd => github.com/containerd/containerd v1.5.0
```
adambkaplan, line 192: Looks like we are pinning containerd to an older version. Can we do one of the following?

- Remove the replace clause here, OR
- Update the containerd version to 1.6.1 so the require and replace versions are in sync.

:crossed_fingers: this avoids the explict replace dependency on imgcrypt.

adambkaplan (changes requested):

go.mod
coreydaley: updated

adambkaplan:
Now I'm confused. I'm looking over the vendor tree, and github.com/containerd/imgcrypt is nowhere to be found.

Unclear how this repo got tagged for [CVE-2022-24778](https://access.redhat.com/security/cve/CVE-2022-24778), this component is not affected.

coreydaley:
I pushed an update to containerd v1.6.2 and replaced containerd/imgcrypt to 1.1.4, I am guessing that v1.6.3 will have the new imgcrypt, so next time we can up to that and remove the replace.

coreydaley:
/close
Since imgcrypt is not a direct dependency

PR #301, opened 2022-06-27
Set _CONTAINERS_USERNS_CONFIGURED if we create a user namespace

nalind:
Set _CONTAINERS_USERNS_CONFIGURED instead of BUILDER_USERNS_CONFIGURED in the environment if we create our own user namespace, so that various bits of logic that depend on detecting that we're in a rootless mode will do what we need them to do.

In this instance, that's setting extended attributes with names starting with "user.\*" instead of "trusted.\*" when applying layers to disk while pulling images.

### DIFF
--- a/cmd/userns.go
+++ b/cmd/userns.go
@@ -14,7 +14,7 @@ import (
 	"k8s.io/klog/v2"
 )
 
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM

PR #304, opened 2022-07-18
bump github.com/containers/buildah to v1.26.2

nalind:
Tweak update-deps.sh to call `go get -t` on the local `cmd` package and the latest buildah package, and pass `-compat=1.17` to `go mod tidy` to avoid an error result.

Update to the current version of buildah, which pulls in v1.40.3 of the storage library, which notably includes https://github.com/containers/storage/pull/1261.

[KATA-1278](https://issues.redhat.com//browse/KATA-1278) is tracking this change for the `buildah` CLI tool, and since we're using it as a library, we're most likely affected by the problem it fixes.  This may not be sufficient for us to avoid problems in that configuration, though.  For cases where we have privileges but can't use kernel overlay, we may need to start trying to create `/dev/fuse` ourselves so that our `fuse-overlayfs` startup-time checks will succeed.


coreydaley:
@nalind 
It looks like just one test is failing.
Would your change cause this?
```
 cgroupv2Val is 
    cat: /sys/fs/cgroup/memory/memory.memsw.limit_in_bytes: No such file or directory
    cgroupv1Val is 
    cat: //sys/fs/cgroup///memory.swap.max: No such file or directory
    cgroupv2Val is 
    cat: /sys/fs/cgroup/cpu,cpuacct/cpu.cfs_quota_us: No such file or directory
    QUOTA=
    cat: /sys/fs/cgroup/cpu,cpuacct/cpu.cfs_quota_us: No such file or directory
    SHARES=
    cat: /sys/fs/cgroup/cpu,cpuacct/cpu.shares: No such file or directory
    PERIOD=
    cat: /sys/fs/cgroup/cpu,cpuacct/cpu.cfs_period_us: No such file or directory
```

nalind:
> @nalind It looks like just one test is failing. Would your change cause this?

I would hope not, but it bears investigating.

PR #305, opened 2022-07-18
Attempt to create /dev/fuse at startup

nalind:
When we're checking if we can use /dev/fuse, if it doesn't exist, try to create it.

When we first added startup-time checks to determine how to set up storage, we didn't foresee being privileged but needing to use fuse-overlayfs anyway.  If the container engine doesn't create `/dev/fuse` for us for whatever reason, that shouldn't prevent us from being able to use it.

This spun out of discussion of KATA-1278 in #304.

### DIFF
--- a/cmd/defaults.go
+++ b/cmd/defaults.go
@@ -2,10 +2,12 @@ package main
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM

PR #306, opened 2022-08-03
bump(github.com/containers/buildah) to v1.26.4

nalind:
* Bump github.com/containers/buildah from v1.26.2 to ~v1.26.3~ v1.26.4, pulling in updates to many of its dependencies.
* Drop a workaround for a bug in v1.26.2 that should be fixed in v1.26.3.
* Use `errors.Is()` to compare error causes to specific errors as some dependencies have switched to using `fmt.Errorf()`'s `%w` verb to do error wrapping, and `errors.Cause()` doesn't unwrap those, and use `errors.As()` in preference to type assertions when checking for specific error types.


coreydaley:
/approve
/assign @rolfedh 
For docs-approved label
/assign @jitendar-singh 
For qe-approved label
/assign @akram @otaviof 
For lgtm

rolfedh (approved):
/lgtm

coreydaley:
/assign @RickJWagner 
for px-approved label

RickJWagner:
Approving with 'px-approved', as PLM has no objection to the provided content.  (Especially happy to see typos in comments cleaned up, etc. just good code maintenance.)

Food for thought:  If we can find a way to break such a large PR into smaller pieces, the reviews can be more targeted (and perhaps of better quality).

coreydaley:
> Approving with 'px-approved', as PLM has no objection to the provided content. (Especially happy to see typos in comments cleaned up, etc. just good code maintenance.)
> 
> Food for thought: If we can find a way to break such a large PR into smaller pieces, the reviews can be more targeted (and perhaps of better quality).

Noted, though the pull request is split into multiple commits here (https://github.com/openshift/builder/pull/306/commits), with dependency bumps we do our best to put the `bump` into one commit, and then any code fixes into a separate (or multiple separate) commits, though with this one we could have combined the two dependency bumps, I think that was just an oversight on my part, I should have asked for them to be merged or to drop the first bump as it was overridden by the second bump.

PR #307, opened 2022-08-15
OCPBUGS-597: bump(github.com/containers/buildah) to v1.29.0

nalind:
Bump github.com/containers/buildah from v1.26.4 to v1.29.0, along with updated versions of a few of its dependencies, ensuring that the primary GID of the user for a RUN command is also in the supplemental groups list (CVE-2022-2990).

Dependencies that previously used github.com/pkg/errors to wrap errors have switched to using the standard library's wrapping facilities, so we can streamline some of our error checking code.

~Add a `RUN --mount=type=bind,from={postCommitStage},source=/run/.containerenv,target=/run/.containerenv :` to the final stage when we add a next-to-final stage to run a `postCommit` command.  The new final image has this added to its history.  Buildah 1.27 skips stages that don't affect the output target stage as an optimization, and the stage that ran the `postCommit` command had no other effect on the final stage; this works around that.~

Use the API setting that 1.27.2 started exposing to let us have it not skip `postCommit` commands.

Update golang.org/x/crypto to ensure we won't be affected by CVE-2022-27191.


nalind:
At least one of the test failures is happening because the post-commit command for an s2i build happens in a Dockerfile stage that doesn't affect the final image, so buildah 1.27 has started skipping the stage.

nalind:
Added a workaround to have the next-to-last stage, if it was added to run the `postCommit` command, run: the new final stage uses `RUN --mount` to bind the `/.containerenv` from the next-to-last stage into the final stage and run `:`, which is a no-op.

nalind:
/retitle bump(github.com/containers/buildah) to v1.27.2

Dropped the workaround for `postCommit` in favor of using an API setting that 1.27.2 started exposing to let us restore the previous behavior.

nalind:
@coreydaley PTAL when it's convenient.

nalind:
/retitle OCPBUGS-597: bump(github.com/containers/buildah) to v1.29.0

v1.26.5 also included the fix for CVE-2022-2990, but let's jump to the current version on this branch, since we generally want to do that anyway.

PR #323, opened 2023-01-19
OCPBUGS-6014: manage-dockerfile: use the original form of HEALTHCHECK

nalind:
When reconstructing the Dockerfile, use the original version of a HEALTHCHECK instruction, rather than incorrectly reassembling all of the arguments into a single JSON array.

### DIFF
--- a/pkg/build/builder/util/dockerfile/dockerfile.go
+++ b/pkg/build/builder/util/dockerfile/dockerfile.go
@@ -40,7 +40,7 @@ func Write(node *parser.Node) []byte {
 				buf.Write(Write(node.Next.Children[0]))
 			}
 			return buf.Bytes()
-		case command.Env, command.Label:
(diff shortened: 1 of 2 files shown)


nalind:
@coreydaley just to set my expectations here, is the e2e-aws-ovn-proxy test expected to pass?

coreydaley:
/hold
@nalind It is not a required test, so I have skipped it. If you are confident that the failures are unrelated to your pull request, please remove the hold on this pull request.

coreydaley:
We were having some api issues with github/jira integration yesterday. I have manually updated the JIRA to the modified state.

Does this need to be backported?

coreydaley:
/assign @jitendar-singh 
for QE

nalind:
This one includes a unit test update, so I think the risk is low, and 4.10's going to be in maintenance phase for several more months, so I FWIW lean toward backporting it.

PR #325, opened 2023-01-30
[release-4.12] Bug 2174485: BUILD-659: bump github.com/containers/buildah to v1.26.6

nalind:
Bump github.com/containers/buildah from v1.26.4 to v1.26.6, ensuring that the primary GID of the user for a RUN command is also in the supplemental groups list (CVE-2022-2990).


nalind:
@coreydaley PTAL

PR #329, opened 2023-02-21
Update CONTRIBUTING.md

divyansh42:


CONTRIBUTING.md
```diff
@@ -35,6 +35,8 @@ $ oc patch clusterversion/version --patch '{"spec":{"overrides":[{"group":"v1",
 ```
 
 2. Make your code changes and build the binary with `make build`.
+> While running `make build` if you come across any error something like
+` OS_GIT_MAJOR: unbound variable`, make sure that you have pulled all the tags from the [openshift/builder](https://github.com/openshift/builder) repo.
```
coreydaley, line 39: I think that we should supply an example of the command to pull the tags

CONTRIBUTING.md
```diff
@@ -54,11 +56,16 @@ $ buildah bud -t <MYREPO>/<MYIMAGE>:<MYTAG> -f Dockerfile.dev .
 $ oc patch configmap openshift-controller-manager-images -n openshift-controller-manager-operator --patch '{"data":{"builderImage":"<MYREPO>/<MYIMAGE>:<MYTAG>"}}' --type=merge
 ```
 
-6. Watch the openshift controller manager daemonset rollout (this can take a few minutes):
+6. Watch the openshift controller manager deployment/daemonset rollout (this can take a few minutes):
 
+If you are running on `4.11` or earlier use:
```
coreydaley, line 61: I am tempted to say that we should remove the part about 4.11, since this doc is for 4.13+, what do you think?

coreydaley (changes requested):

CONTRIBUTING.md
divyansh42: Make sense :+1:

CONTRIBUTING.md
divyansh42: Done

PR #330, opened 2023-02-28
OCPBUGS-3064: normalize locations in pull/push secrets which might look like URLs

nalind:
Normalize locations in secrets that look like URLs, since we used to accept (or "accept") those.

Updates mergeNodeCredentials(), used by the image pulling logic. Updates mergeNodeCredentialsDockerAuth(), used by the docker and sti strategies. Updates extractSourceFromImage() directly, since it doesn't use either of the other functions.

### DIFF
--- a/pkg/build/builder/common.go
+++ b/pkg/build/builder/common.go
@@ -11,6 +11,7 @@ import (
 	"os"
 	"path/filepath"
 	"sort"
+	"strconv"
 	"strings"
 	"time"
 
@@ -487,12 +488,10 @@ func replaceImagesFromSource(node *parser.Node, imageSources []buildapiv1.ImageS
 				}
 				names[stage.Name] = image
 			case child.Value == dockercmd.Copy:
-				if ref, ok := nodeHasFromRef(child); ok {
-					if len(ref) > 0 {
-						if _, ok := names[ref]; !ok {
-							if replacement, ok := replacements[ref]; ok {
-								nodeReplaceFromRef(child, replacement)
-							}
+				if ref, ok := nodeHasFromRef(child); ok && len(ref) > 0 {
+					if _, ok := names[ref]; !ok {
+						if replacement, ok := replacements[ref]; ok {
+							nodeReplaceFromRef(child, replacement)
 						}
 					}
 				}
@@ -511,9 +510,15 @@ func findReferencedImages(dockerfilePath string, buildArgs []corev1.EnvVar) ([]s
 	if err != nil {
 		return nil, err
 	}
-	names := make(map[string]string)
 	images := sets.NewString()
+	headerArgs := dockerfile.HeaderArgs(node)
 	buildArgMap := make(map[string]string)
+	for _, arg := range headerArgs {
+		kv := strings.SplitN(arg, "=", 2)
+		if len(kv) == 2 {
+			buildArgMap[kv[0]] = kv[1]
+		}
+	}
 	for _, ba := range buildArgs {
 		buildArgMap[ba.Name] = ba.Value
 	}
@@ -521,25 +526,38 @@ func findReferencedImages(dockerfilePath string, buildArgs []corev1.EnvVar) ([]s
 	if err != nil {
 		return nil, err
 	}
-	for _, stage := range stages {
+	var argsSlice []string
+	for k, v := range buildArgMap {
+		argsSlice = append(argsSlice, fmt.Sprintf("%s=%s", k, v))
+	}
+	knownAliases := make(map[string]struct{})
+	for nthStage, stage := range stages {
 		for _, child := range stage.Node.Children {
 			switch {
 			case child.Value == dockercmd.From && child.Next != nil:
-				image := child.Next.Value
-				names[stage.Name] = image
-				if _, ok := names[image]; !ok {
+				image, err := imagebuilder.ProcessWord(child.Next.Value, argsSlice)
+				if err != nil {
+					return nil, err
+				}
+				if _, alias := knownAliases[image]; !alias {
 					images.Insert(image)
 				}
+				if child.Next != nil && child.Next.Next != nil && strings.ToUpper(child.Next.Next.Value) == "AS" && child.Next.Next.Next != nil {
+					knownAliases[child.Next.Next.Next.Value] = struct{}{}
+				}
 			case child.Value == dockercmd.Copy:
-				if ref, ok := nodeHasFromRef(child); ok {
-					if len(ref) > 0 {
-						if _, ok := names[ref]; !ok {
-							images.Insert(ref)
-						}
+				if ref, ok := nodeHasFromRef(child); ok && len(ref) > 0 {
+					image, err := imagebuilder.ProcessWord(ref, argsSlice)
+					if err != nil {
+						return nil, err
+					}
+					if _, alias := knownAliases[image]; !alias {
+						images.Insert(image)
 					}
 				}
 			}
 		}
+		knownAliases[strconv.Itoa(nthStage)] = struct{}{}
 	}
 	return images.List(), nil
 }
--- a/pkg/build/builder/util.go
+++ b/pkg/build/builder/util.go
@@ -166,3 +166,49 @@ func NameForBuildVolume(objName string) string {
 func PathForBuildVolume(objName string) string {
 	return filepath.Join(buildVolumeMountPath, NameForBuildVolume(objName))
 }
+
+// normalizeRegistryLocation munges a registry location that's mistakenly been
+// provided in the old http/https URL format, which containers/image now
+// rejects, into a proper location prefix.  Locations which don't look like
+// http: or https: URLs are returned unmodified.
+//
+// URL path components are preserved, which is a change from how this would
+// have worked in earlier 4.x, but brings behavior closer to how pushing
+// credentials are handled (and how this worked in 3.x): those credentials are
+// selected by matching the registry name using filename-style matching of the
+// host name component, and by checking if the path component is a string (not
+// path) prefix of the repository that the image will be pushed to, as is
+// typical for Kubernetes.
+//
+// The callers of this function are using its result to supply credentials to
+// containers/image to select from while pulling an image.
+// containers/image doesn't implement the wildcard or string-prefix matching
+// logic when it's deciding which of the credentials it's been passed should be
+// used, and in older versions which accepted http: and https: URLS, it ignored
+// the path component, so in addition to letting the builder accept them again,
+// this allows path components in URLs to be matched using container/image
+// rules (i.e., no wildcard and prefix matching).
+//
+// An alternate approach would have been to populate a
+// k8s.io/kubernetes/pkg/credentialprovider.BasicDockerKeyring with all of the
+// secrets we know, pass the name of every image we'll pull to its Lookup()
+// method, and pass the image library the returned credential for each image's
+// location.  The challenge of predicting which images will be pulled along
+// every code path, and factoring in search registries, made this approach more
+// attractive.
+func normalizeRegistryLocation(location string) string {
+	if !strings.HasPrefix(location, "http://") && !strings.HasPrefix(location, "https://") {
+		return location
+	}
+	cleaned := strings.Split(strings.TrimPrefix(strings.TrimPrefix(location, "http://"), "https://"), "/")
+	switch cleaned[0] {
+	case "index.docker.io", "registry-1.docker.io":
+		cleaned[0] = "docker.io"
+	}
+	if len(cleaned) > 1 {
+		if cleaned[1] == "v1" || cleaned[1] == "v2" {
+			cleaned = append(cleaned[:1], cleaned[2:]...)
+		}
+	}
+	return strings.TrimSuffix(strings.Join(cleaned, "/"), "/")
+}
--- a/pkg/build/builder/common_test.go
+++ b/pkg/build/builder/common_test.go
@@ -637,6 +637,24 @@ func Test_findReferencedImages(t *testing.T) {
 				Images: []string{"other", "scratch", "test"},
 			},
 		},
+		{
+			original: heredoc.Doc(`
+				ARG headerArg=h
+				FROM other:${headerArg} AS alias
+				ARG middleArg=h
+				COPY --from=testImage /a /b
+				FROM alias as testStage
+				COPY --from=testImage /a /b
+				FROM other:${unsetArg}
+				COPY --from=testStage /a /b
+				FROM other:${headerArg}
+				COPY --from=3 /a /b
+				COPY --from=2 /a /b
+				`),
+			want: want{
+				Images: []string{"3", "other:", "other:h", "testImage"},
+			},
+		},
 	}
 	for i, test := range tests {
 		t.Run(fmt.Sprintf("%d", i), func(t *testing.T) {
--- a/pkg/build/builder/daemonless.go
+++ b/pkg/build/builder/daemonless.go
@@ -112,30 +112,30 @@ func mergeNodeCredentials(credsPath string) *credentialprovider.DockerConfigJSON
 	if err != nil {
 		log.V(2).Infof("proceeding without build pull credentials: %v", err)
 	}
-	if nodeCreds == nil || len(nodeCreds) == 0 {
-		return &credentialprovider.DockerConfigJSON{
-			Auths: pullCreds,
-		}
-	}
-	if pullCreds == nil || len(pullCreds) == 0 {
-		return &credentialprovider.DockerConfigJSON{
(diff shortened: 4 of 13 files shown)


pkg/build/builder/util.go
```diff
@@ -166,3 +166,22 @@ func NameForBuildVolume(objName string) string {
 func PathForBuildVolume(objName string) string {
 	return filepath.Join(buildVolumeMountPath, NameForBuildVolume(objName))
 }
+
+// normalizeRegistryLocation munges a registry location that's mistakenly been
+// provided in the old http/https URL format into a proper location prefix
+func normalizeRegistryLocation(location string) string {
```
nalind, line 172: @mtrmac could I trouble you for a once-over on this one?  

It's not as clean as https://github.com/containers/image/blob/4fcb52359b759a276a760fb28c939c357b3aea48/pkg/docker/config/config.go#L732, but it's trying to be compatible with the examples listed in https://access.redhat.com/documentation/en-us/openshift_container_platform/4.10/html-single/cicd/index#builds-docker-credentials-private-registries_creating-build-inputs, which include namespaces in the URLs, while we work on updating the docs.

nalind: (Just this function, and maybe the test cases below - the rest you don't need to worry about unless you're _really_ curious about this.)

mtrmac: A quick check / some references for now, I can investigate in more detail tomorrow if you’d like me to.

Primarily, is this supposed to be based on the Podman semantics or the Kubernetes semantics? Is the only reference the linked documentation? Does none of that matter, and do we get to _choose_ what to implement, or are we strongly constrained?
  - (Would it be plausible to change only the docs? The semantic ambiguity of `…/v2` is really not elegant.)
  - Podman config file semantics: “Repo or namespace keys are only supported as exact matches” (i.e. not in URL form); URLs are only interpreted as whole-registry credentials (so they should be normalized to non-URL host[:port]).
  - Kubernetes semantics: `v1` and `v2` special-cased, https://github.com/kubernetes/kubernetes/blob/18eea58ac2da8186b6315c90db9a5cc11048c69d/pkg/credentialprovider/keyring.go#L105 and wildcard matches: https://github.com/kubernetes/kubernetes/blob/18eea58ac2da8186b6315c90db9a5cc11048c69d/pkg/credentialprovider/keyring.go#L216 . Compare also https://github.com/kubernetes/kubernetes/issues/102472 .
  - (`Podman login` semantics: https://github.com/containers/common/blob/18c4568e8ee051fd8ded1d9d47cfa0453b1d6a0c/pkg/auth/auth.go#L188  _probably_ irrelevant for this input format.)

The reference to “pull secrets” suggests that this might be expected to be the same as the k8s `kubernetes.io/dockerconfigjson` format, with presumably exactly the same semantics — OTOH (per https://github.com/kubernetes/kubernetes/issues/102472#issuecomment-975748695 ) I’d really prefer _not_ to spread the Kubelet’s semantics format any further, if _reasonably possible_.

I guess (without having carefully read the code or the report) if the goal is only to find duplicates, this code doesn’t really need to take a position on the K8s wildcard and prefix semantics at all… — but the difference in how K8s and Podman treats URLs is material for normalization/deduplication; at least that part needs to be decided.

mtrmac: Assuming the change in `buildDaemonlessImage` is binding, i.e. that URLs were previously completely rejected, it does seem attractive to me to refuse to support both of the ambiguous formats.

Also, c/image (and, I think, Docker) ultimately doesn’t have any support for the Kubernetes wildcard or prefix matching, so if the credentials end up being processed by Buildah or Docker, the full Kubernetes semantics is problematic.

nalind: Wow, that is... a lot to digest.

The docs are being updated, but I don't have a timeline for that, and in the meantime I think we've been producing errors after upgrades.

The OpenShift documentation linked above includes an example of a config file being imported as a secret with type `kubernetes.io/dockerconfigjson`, so that seems to be the expectation that's been set.

At the points this PR touches, though, it's mainly concerned with passing secret contents in as API structures and not triggering an error, hopefully by approximating what a human being would do if they were to edit the secret directly.  The selection of which credential of possibly-many that we pass in should be used is still down in the image library.  I'm not sure we want to go there.

mtrmac: Organic growth in the 4 or so parallel implementations has not been kind to us.

---

>  in the meantime I think we've been producing errors after upgrades.

Oh, so there is, _also_, a compatibility constraint for openshift/builder? Is that triggered by a c/image behavior change, or by something in openshift/builder?

If it is c/image (I didn’t, so far, look into the openshift/builder history to know the details), I guess it is c/image 2935f06f71d10b7c6961d6a70d2ff428d3fdbde3 — before that, all URLs were treated as whole-registry credentials, not restricted to namespaces/repos . (In fact the namespace matching was _only_ added a few days before that, c/image 2cafc0105c89b1c569f68733f7a5eaf6e326831b ; and evolved a bit afterwards.)

---

> The selection of which credential of possibly-many that we pass in should be used is still down in the image library. I'm not sure we want to go there.

I’m also tempted to prefer not to see things :) https://github.com/openshift/openshift-controller-manager/blob/master/pkg/build/controller/build/build_controller.go , and, even, https://github.com/openshift/builder/blob/master/pkg/build/builder/cmd/dockercfg/cfg.go , seem to use the Kubernetes semantics. It has never worked that way on the c/image paths, and at the very least least the prefix matching using `quay.io/hello` credentials for `quay.io/hello-openshift` is something I’d prefer not to implement.

But, of course, business needs and customer promises do matter…

----

If I mostly ignore the question of cross-product semantics consistency; and focus _only_ on the code paths that involve c/image; and if I assume that the goal is to navigate a transition across c/image adding per-namespace keys and enforcing strict validation for `SetCredentials` (per the unverified _guess_ above, as it happened in 2021):

- `host.com` and `host.com:port` should not be normalized in any way, apart from maybe the three `docker.io` host names (where it doesn’t matter too much how they are normalized , except that builder’s deduplication requires the normalization to be consistent).
- `host.com/any-namespace/sequence` should not be normalized _at all_; for Docker Hub the host name should be exactly `docker.io`, not one of the two others.
- `http://…` and `https://…` are, by c/image, treated as whole-registry credentials, so as far as c/image is concerned, any normalization of the URL path part does not matter. **But turning `http[s]://host.com/namespace` into `host.com/namespace` changes semantics.**

So the two “easy” options seem to be:
- Turn `http[s]://` URLs into per-namespace credentials, by stripping the scheme and keeping the rest. Breaks compatibility with the old behavior of openshift/builder, before it started rejecting URLs: URLs were interpreted by c-image as whole-registry scopes; but it interprets URLs similarly (but not _the same_) as the Kubelet. In some (the “easy and reasonable”) cases c/image and Kubelet will interpret the scope the same way.
- `turn `http[s]://` URLs into whole-registry credentials, by turning them into host[:port] only. AFAICS compatible with the old behavior of openshift/builder, before it started rejecting URLs, but also _always_ interprets the scope differently from Kubelet.

nalind: > Organic growth in the 4 or so parallel implementations has not been kind to us.
> 
> > in the meantime I think we've been producing errors after upgrades.
> 
> Oh, so there is, _also_, a compatibility constraint for openshift/builder? Is that triggered by a c/image behavior change, or by something in openshift/builder?

Of containers/image and the logic in openshift/builder that takes the secret files and passes them in to the APIs, I guess containers/image changed more recently, if that's what you're asking.

I took another look, and we started including the https:// form in an example when the CI/CD overview doc was introduced in 4.7, and expanded upon it by adding examples that included namespaces in 4.9.  Before that I don't know for certain that we implied that it was accepted.  We have a customer report that the build threw out an error out when they tried it, and it does appear to be something that would have been accepted in 3.x that I didn't note as a behavior change during the move to 4.x.

> If it is c/image (I didn’t, so far, look into the openshift/builder history to know the details), I guess it is c/image 2935f06f71d10b7c6961d6a70d2ff428d3fdbde3 — before that, all URLs were treated as whole-registry credentials, not restricted to namespaces/repos . (In fact the namespace matching was _only_ added a few days before that, c/image 2cafc0105c89b1c569f68733f7a5eaf6e326831b ; and evolved a bit afterwards.)
> 
> > The selection of which credential of possibly-many that we pass in should be used is still down in the image library. I'm not sure we want to go there.
> 
> I’m also tempted to prefer not to see things :) https://github.com/openshift/openshift-controller-manager/blob/master/pkg/build/controller/build/build_controller.go , and, even, https://github.com/openshift/builder/blob/master/pkg/build/builder/cmd/dockercfg/cfg.go , seem to use the Kubernetes semantics.

Oh, I'd overlooked those.  Or forgotten about them.

> If I mostly ignore the question of cross-product semantics consistency; and focus _only_ on the code paths that involve c/image; and if I assume that the goal is to navigate a transition across c/image adding per-namespace keys and enforcing strict validation for `SetCredentials` (per the unverified _guess_ above, as it happened in 2021):
> 
> * `host.com` and `host.com:port` should not be normalized in any way, apart from maybe the three `docker.io` host names (where it doesn’t matter too much how they are normalized , except that builder’s deduplication requires the normalization to be consistent).
> 
> * `host.com/any-namespace/sequence` should not be normalized _at all_; for Docker Hub the host name should be exactly `docker.io`, not one of the two others.

I'm not clear on the distinction between "in any way" and "not at all".  Were you saying that one case or the other should generate an error, or were you just switching up the word choice to avoid them sounding repetitive?

> * `http://…` and `https://…` are, by c/image, treated as whole-registry credentials, so as far as c/image is concerned, any normalization of the URL path part does not matter. **But turning `http[s]://host.com/namespace` into `host.com/namespace` changes semantics.**

Those semantics are closer to what the build controller and the parts of the builder that push images do, as you pointed out.  Hmm.

> So the two “easy” options seem to be:
> 
> * Turn `http[s]://` URLs into per-namespace credentials, by stripping the scheme and keeping the rest. Breaks compatibility with the old behavior of openshift/builder, before it started rejecting URLs: URLs were interpreted by c-image as whole-registry scopes; but it interprets URLs similarly (but not _the same_) as the Kubelet. In some (the “easy and reasonable”) cases c/image and Kubelet will interpret the scope the same way.
> * Turn `http[s]://` URLs into whole-registry credentials, by turning them into host[:port] only. AFAICS compatible with the old behavior of openshift/builder, before it started rejecting URLs, but also _always_ interprets the scope differently from Kubelet.

The first of these options makes parts of openshift/builder more consistent with earlier versions of itself (give or take the prefix and wildcard matching, which I am happy to let containers/image continue to not do), and makes pushing and pulling in the builder more consistent with each other.  I _think_ it's what we get with the PR in its current form, too, so I'm leaning that way.

Does that reasoning sound... reasonable, or have I lost the thread here?

mtrmac: > I'm not clear on the distinction between "in any way" and "not at all"

My fault for sloppy wording, I didn’t mean anything specific by that — the “not at all” wording was just to emphasize the `docker.io` (non-)normalization discussed afterwards.

(c/image wouldn’t refuse to process files with e.g. `index.docker.io/library`, but it would never match that entry. But `SetCredentials` also does not reject requests to create that… probably should be fixed.)

---

Yes, I think turning `https://example.com/foo` not `example.com/foo` is a very reasonable choice. Both of the two alternatives have broken/surprising behavior in _some_ cases, and the uses of `keyring.Lookup` elsewhere in this repo is, I think, a good reason to use this one.

---

> I think it's what we get with the PR in its current form, too, so I'm leaning that way.

One nit: I’m not sure what `normalizeRegistryLocation` should do about `{index,registry-1}.docker.io` with a path part, e.g. `index.docker.io/foo` and `http[s]://index.docker.io/foo`. 

In the c/image semantics, both `[https://]index.docker.io/foo` mean `docker.io` (the whole registry) - with a scheme because scheme entries are defined the whole registry, and without a scheme because non-canonical `*.docker.io` are not found as namespace matches and fall into the compatibility path.

Looking at https://github.com/kubernetes/kubernetes/blob/b6d102d634d357600bc0fcb076395c88caef6acb/pkg/credentialprovider/keyring.go#L254-L255 in the K8s semantics, users can also have _either_ wildcard and prefix matches, _or_ `*.docker.io` normalization. (I didn’t actually test that now.)

And I could see an argument that the user “clearly meant” for `index.docker.io/foo` to match `docker.io/foo`, so I don‘t feel strongly about it; OTOH if there are multiple secrets, matching them more similarly to the Kubelet semantics seems worthwhile.

---

I think it would be useful to fairly extensively document the various considerations in the final version of the code.

coreydaley:
Error creating: pods "aws-ebs-csi-driver-node-" is forbidden: unable to validate against any security context constraint: [provider "anyuid": Forbidden: not usable by user or serviceaccount, provider restricted-v2: .spec.securityContext.hostNetwork: Invalid value: true: Host network is not allowed to be used, spec.volumes[0]: Invalid value: "hostPath": hostPath volumes are not allowed to be used, spec.volumes[1]: Invalid value: "hostPath": hostPath volumes are not allowed to be used, spec.volumes[2]: Invalid value: "hostPath": hostPath volumes are not allowed to be used, spec.volumes[3]: Invalid value: "hostPath": hostPath volumes are not allowed to be used, spec.volumes[4]: Invalid value: "hostPath": hostPath volumes are not allowed to be used, spec.volumes[5]: Invalid value: "hostPath": hostPath volumes are not allowed to be used, spec.containers[0].securityContext.privileged: Invalid value: true: Privileged containers are not allowed, spec.containers[0].securityContext.hostNetwork: Invalid value: true: Host network is not allowed to be used, spec.containers[0].securityContext.containers[0].hostPort: Invalid value: 10300: Host ports are not allowed to be used, spec.containers[1].securityContext.privileged: Invalid value: true: Privileged containers are not allowed, spec.containers[1].securityContext.hostNetwork: Invalid value: true: Host network is not allowed to be used, spec.containers[1].securityContext.containers[0].hostPort: Invalid value: 10300: Host ports are not allowed to be used, spec.containers[2].securityContext.hostNetwork: Invalid value: true: Host network is not allowed to be used, spec.containers[2].securityContext.containers[0].hostPort: Invalid value: 10300: Host ports are not allowed to be used, provider "restricted": Forbidden: not usable by user or serviceaccount, provider "nonroot-v2": Forbidden: not usable by user or serviceaccount, provider "nonroot": Forbidden: not usable by user or serviceaccount, provider "hostmount-anyuid": Forbidden: not usable by user or serviceaccount, provider "machine-api-termination-handler": Forbidden: not usable by user or serviceaccount, provider "hostnetwork-v2": Forbidden: not usable by user or serviceaccount, provider "hostnetwork": Forbidden: not usable by user or serviceaccount, provider "hostaccess": Forbidden: not usable by user or serviceaccount, provider "privileged": Forbidden: not usable by user or serviceaccount] for DaemonSet.apps/v1/aws-ebs-csi-driver-node -n openshift-cluster-csi-drivers happened 11 times

pkg/build/builder/util.go
nalind: Took a pass at it.

nalind:
Documenting considerations for `normalizeRegistryLocation` turned up places where `findReferencedImages` wasn't quite doing the right thing.  Made the latter a bit more thorough, and expanded the unit tests.

pkg/build/builder/common.go
```diff
@@ -488,7 +489,7 @@ func replaceImagesFromSource(node *parser.Node, imageSources []buildapiv1.ImageS
 				names[stage.Name] = image
 			case child.Value == dockercmd.Copy:
 				if ref, ok := nodeHasFromRef(child); ok {
-					if len(ref) > 0 {
+					if ok && len(ref) > 0 {
```
mtrmac, line 492: (I didn’t really read the new code — just skimming to see what is affected.

`ok` was checked one line above; this condition could be folded there.)

nalind: Oof, right.

coreydaley:
/retest
infrastructure issue provisioning resources

nalind:
/assign @coreydaley 
@mtrmac I think I captured the how-we-got-here in a block comment [here](https://github.com/openshift/builder/pull/330/files#diff-558b6bd45753530119c7b462bb0ac697d81a6843623ddffc1fa893acb2122511R170), but if I've left anything out, please remind me of it.

mtrmac:
> @mtrmac I think I captured the how-we-got-here in a block comment [here](https://github.com/openshift/builder/pull/330/files#diff-558b6bd45753530119c7b462bb0ac697d81a6843623ddffc1fa893acb2122511R170), but if I've left anything out, please remind me of it.

I think that’s accurate and comprehensive.

coreydaley:
I would like it to pass the tests again and then I think we can merge it
/lgtm
/approve

PR #335, opened 2023-04-04
OCPBUGS-7782: bump(github.com/containers/common) to v0.51.2

nalind:
Bump github.com/containers/common from v0.51.0 to v0.51.2, which puts CAP_SYS_CHROOT back into the default list of capabilities that we grant when handling RUN instructions.


nalind:
Opened #336 to backport this change to the "release-4.13" branch.

PR #340, opened 2023-04-06
OCPBUGS-1973: Pass CPU limits to the container

coreydaley:
It looks like S2I was already doing this, it just never got implemented in Docker and Daemonless

https://github.com/openshift/builder/blob/master/pkg/build/builder/sti.go#L359-L366

### DIFF
--- a/pkg/build/builder/daemonless.go
+++ b/pkg/build/builder/daemonless.go
@@ -316,6 +316,9 @@ func buildDaemonlessImage(sc types.SystemContext, store storage.Store, isolation
 		},
 		CommonBuildOpts: &buildah.CommonBuildOptions{
 			HTTPProxy:          true,
+			CPUPeriod:          uint64(opts.CPUPeriod),
+			CPUShares:          uint64(opts.CPUShares),
+			CPUQuota:           opts.CPUQuota,
 			Memory:             opts.Memory,
 			MemorySwap:         opts.Memswap,
 			CgroupParent:       opts.CgroupParent,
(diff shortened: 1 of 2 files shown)


coreydaley:
/assign @nalind 
Would you mind taking a look at this for me and make sure that I am not doing anything dumb? :)

coreydaley:
/assign @bparees 
ptal if you get a chance.

coreydaley:
/retest
infra issues

bparees:
i'm also interested in nalin's take.  my expectation is that this is safe because we were already cpu throttling due to running the container in the parent cgroup, so nothing should really change for the build process other than that the cpu cgroup values are explicilty exposed instead of just implicity enforced.

also iirc we have tests that ensure the memory limits are seen within the build, you should update those tests to do the same for cpu limits as part of this.

coreydaley:
/hold
It looks like the CPU Limits may not be getting populated correctly.

nalind:
There might still be a missing piece or two.  If I'm reading `builderConfig.execute()` in `pkg/build/builder/cmd/builder.go` right, some of the cgroup settings that we'll be applying are retrieved from the builder process's cgroups by calling `GetCGroupLimits()` in `pkg/build/builder/util_linux.go`, which currently doesn't appear to be reading CPU-related limits.  Those settings are then passed down to `DockerBuilder.dockerBuild()` in `pkg/build/builder/docker.go`.  So `GetCGroupLimits()` might need an update as well.

I'd be tempted to try forcing everything to use the "host"'s cgroup kernel namespace, which would let us simplify the code by removing a bunch of it.

coreydaley:
/close
in favor of #341

PR #341, opened 2023-04-12
OCPBUGS-1973: pass CPU limits for Docker strategy builds

nalind:
Teach `pkg/build/builder.GetCGroupLimits()` to read cgroups v1 "cpu.shares", "cpu.cfs_quota_us", and "cpu.cfs_period_us" files, and to read more or less equivalent values from v2 "cpu.max" and "cpu.weight" files.

When reading v2 values, scale the "cpu.weight" from the v2 [1..10000] weight range to the v1 [2..0x40000] share range, and scale the "cpu.max" values from the v2 [0..100000] range to the v1 [0..1000000] ranges.

Pick up @coreydaley's changes from #340 to set CPU limits when using the Docker strategy.

### DIFF
--- a/pkg/build/builder/daemonless.go
+++ b/pkg/build/builder/daemonless.go
@@ -316,6 +316,9 @@ func buildDaemonlessImage(sc types.SystemContext, store storage.Store, isolation
 		},
 		CommonBuildOpts: &buildah.CommonBuildOptions{
 			HTTPProxy:          true,
+			CPUPeriod:          uint64(opts.CPUPeriod),
+			CPUShares:          uint64(opts.CPUShares),
+			CPUQuota:           opts.CPUQuota,
 			Memory:             opts.Memory,
 			MemorySwap:         opts.Memswap,
 			CgroupParent:       opts.CgroupParent,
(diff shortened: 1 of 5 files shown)


coreydaley:
/hold
while I look for tests that need to be updated

nalind:
I'd want to extend the s2i-with-quota tests to check for CPU quota settings (for the sake of cgroups v2, they'd have to have a tolerance for rounding errors caused by using integer math, or we'll have to use floating point numbers for some of the scaling we do to convert the v2 values we read from the kernel into values that are suitable for setting in a runtime config, since the data structure expects to include v1 values, which get converted back into v2 values when the crun/runc we invoke notice that the host is using cgroups v2), and add tests that correspond to all of the s2i-with-quota tests, but using the docker strategy.

coreydaley:
@nalind That sounds like a good plan to me. Are you up to adding the tests? We can test the tests (weird right?) against this pull request using cluster bot.

nalind:
@coreydaley https://github.com/openshift/origin/pull/27891 includes the tests I think we want for this one.

nalind:
@coreydaley I think we're about ready to merge this - the last set of changes to https://github.com/openshift/origin/pull/27891 made the tests it touches pass (the run as a whole did not, YMMV), and I've [started a fresh run](https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/release-openshift-origin-installer-launch-aws-modern/1658088502312570880) today to get another sample.

PR #349, opened 2023-06-05
BUILD-432: mount secrets as "overlay" mounts

nalind:
Mount secrets as "overlay" mounts, so that we can more easily mark them as noexec/nosuid/nodev in unprivileged builds, where attempting to change the mount flags on a bind mount in an unprivileged namespace would trigger errors.

Log which capabilities we have at startup at level 2 or higher, and use the newer API to check if we have CAP_SYS_ADMIN.


rhatdan:
LGTM
@flouthoc @giuseppe @vrothberg @TomSweeneyRedHat @umohnani8 PTAL

nalind:
@coreydaley PTAL

nalind:
This is part of BUILD-432, but not necessarily all of it.

PR #358, opened 2023-09-13
OCPBUGS-23128: bump github.com/containers/buildah to fix transient mounting in chroot isolation

nalind:
* bump github.com/containers/buildah from v1.32.0 to v1.32.1 to get improvements in how its chroot logic handles bind mounts when the read-only mount option is combined with nodev/noexec/nosuid being used at the source of the bind mount
* wire up klog's CLI flag parsing to main, making the old `--loglevel` option a proper alias for its `--v` option, and exposing klog's other flags
* also log when we're started with node-default user namespace mappings, so that tests of unprivileged builds will be able to tell the difference between running in a user namespace (the log message we already emitted at level 2), not running in a user namespace (a new log message at level 2), or a problem turning up the logging level (which was previously indistinguishable from the not-running-in-a-user-namespace case, since they both involved not logging ID mapping information)
* defer copying CA certificates supplied by the build controller to locations where the libraries look for them until after we've processed CLI flags, so that `--help` doesn't always try to modify things on disk

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -45,23 +45,31 @@ func main() {
 		runtime.GOMAXPROCS(runtime.NumCPU())
 	}
 
-	const tlsCertRoot = "/etc/pki/tls/certs"
-	const runtimeCertRoot = "/etc/docker/certs.d"
+	installCAcerts := func() {
+		const tlsCertRoot = "/etc/pki/tls/certs"
+		const runtimeCertRoot = "/etc/docker/certs.d"
 
-	clusterCASrc := fmt.Sprintf("%s/ca.crt", builder.SecretCertsMountPath)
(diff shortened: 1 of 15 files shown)


nalind:
I don't want to have to work around not having https://github.com/openshift/origin/pull/28352, but we can go back to having `RunGitClone()` omit its name in its log message if we have to.

coreydaley:
@nalind This needs a bug in OCPBUGS

nalind:
/title OCPBUGS-23128: bump github.com/containers/buildah to fix transient mounting in chroot isolation

nalind:
/retitle OCPBUGS-23128: bump github.com/containers/buildah to fix transient mounting in chroot isolation

PR #359, opened 2023-10-09
Drop the Overlay setting from transient mounts

nalind:
Drop the Overlay flag from the set of flags that we pass in for our transient mounts.

We erroneously assumed that it was safe to set the flag on all secrets, but it can't be used, in the sense that we expect it to, when the item to be transiently mounted is not a directory.

We'll need to backport to 4.14 once we merge this.


nalind:
@coreydaley PTAL

rhatdan:
LGTM

PR #392, opened 2024-05-15
OCPBUGS-25495: Upgrade s2i and buildah

adambkaplan:
Upgrade s2i and buildah to latest versions. These collectively help mitigate CVE-2023-48795


adambkaplan:
@nalind this is interesting - builds aren't able to pull the base image. Seeing the following message:

```
could not find "netavark" in one of {[/usr/local/libexec/podman /usr/local/lib/podman /usr/libexec/podman /usr/lib/podman] {<nil>}}.  To resolve this error, set the helper_binaries_dir key in the `[engine]` section of containers.conf to the directory containing your helper binaries.
  2024-05-15T20:38:13.308570874Z Warning: Pull failed, retrying in 5s ...
```

nalind:
There's a check for the default networking backend that we have to do pretty early due to the way the backward compatibility logic for using CNI was implemented.  The builder ends up using "host" (in the container) networking when handling RUN instructions, but the dependency is still there because the library doesn't know that we'll never ask for it.

adambkaplan:
ack - given that it is part of Podman 4 and can be installed on UBI9 via dnf, adding that package to the set of dependencies is fine by me. But I won't want it to land in 4.16 just yet - at this point bumping buildah should merge for 4.17 so the update gets plenty of "soak time." We'll only backport the golang/x/crypto dep for 4.16 and earlier.

adambkaplan:
/label px-approved

No impact to product experience

/label docs-approved

Release note will provide documentation

PR #394, opened 2024-05-29
OCPBUGS-33215: Bumps openshift/api & related dependencies

apoorvajagtap:
The current vendored version of openshift/api doesn't include the changes merged with https://github.com/openshift/api/pull/1674. This commit bumps the openshift/api to the latest release-4.15 branch.


sayan-biswas (approved):

go.mod
```diff
@@ -24,11 +24,11 @@ require (
 	github.com/spf13/pflag v1.0.5
 	github.com/syndtr/gocapability v0.0.0-20200815063812-42c35b437635
 	golang.org/x/sys v0.18.0
-	k8s.io/api v0.28.2
-	k8s.io/apimachinery v0.28.2
+	k8s.io/api v0.30.1
+	k8s.io/apimachinery v0.30.1
```
adambkaplan, line 28: Did you mean to bump k8s here?

All core k8s.io dependencies need to be on the same version. We have a replace stanza in the go mod that does this, since we do a "bad thing" and import kubernetes/kubernetes directly.

vendor/modules.txt
```diff
@@ -1162,7 +1161,7 @@ gopkg.in/yaml.v2
 # gopkg.in/yaml.v3 v3.0.1
 ## explicit
 gopkg.in/yaml.v3
-# k8s.io/api v0.28.2 => k8s.io/api v0.28.2
+# k8s.io/api v0.30.1 => k8s.io/api v0.28.2
```
adambkaplan, line 1164: Down here go does the right thing and downgrades k8s.

We should upgrade k8s in a follow up PR for OCP 1.17.

adambkaplan (commented):
If possible, please revert the core k8s version upgrades.

go.mod
apoorvajagtap: @adambkaplan actually bumping to github.com/openshift/api@release-4.16 led to k8s.io/api version > 0.28.2.
As our current requirement is to include the updated [whitelist envs](https://github.com/openshift/api/pull/1674/files), I have used now github.com/openshift/api@release-4.15 version atm. This keeps the k8s.io dependencies on v0.28.2. PTAL

PR #411, opened 2024-11-26
OCPBUGS-44243:unit test issue fix on git submodule creation

prabhapa:
This is the test PR to fix the bug OCPBUGS-44243 . Unit test case that create git submodule fails permanently. This is a test to see if it fixes.


prabhapa:
this is a test hence closing

PR #412, opened 2024-12-04
OCPBUGS-45321: skipping some unit tests to avoid failures as they are duplicate

prabhapa:
Recent git upgrade to fix a CVE is causing some issues lately for the unit tests. The git upgrade disabled file protocol by default which is causing the problem.Those tests are skipped as they are duplicate.

### DIFF
--- a/pkg/build/builder/source_test.go
+++ b/pkg/build/builder/source_test.go
@@ -162,6 +162,7 @@ func (r *testGitRepo) addCommit() error {
 }
 
 func TestUnqualifiedClone(t *testing.T) {
+	t.Skip("OCPBUGS-45321: git file:// protocol no longer allowed by default. See CVE-2022-39253")
 	repo, err := initializeTestGitRepo("unqualified")
 	defer repo.cleanup()
 	if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/build/builder/source_test.go
```diff
@@ -162,6 +162,7 @@ func (r *testGitRepo) addCommit() error {
 }
 
 func TestUnqualifiedClone(t *testing.T) {
+	t.Skip("This test as its failing permanently due to Git upgrade after a CVE.Similar tests happen on other places hence skipping it")
```
adambkaplan, line 165: Be specific, and reference JIRA or other identifiers where appropriate. Ditto for the other comments.

```suggestion
	t.Skip("OCPBUGS-45321: git file:// protocol no longer allowed by default. See CVE-2022-39253")
```

adambkaplan (changes requested):
In addition to the updated skip message, please use a well-formatted and _specific_ git commit message.

I love referencing the Tekton community standards, because they do an excellent job of explaining what a good commit message is: https://github.com/tektoncd/community/blob/main/standards.md#commit-messages

adambkaplan:
/approve

Acceptable to skip the unit tests due to change in git's behavior. We have plenty of git test coverage via the end to end tests in openshift/origin.

adambkaplan:
/override ci/prow/e2e-aws-ovn-builds

Known issue with builds e2e test

adambkaplan:
Please squash commits, then I can add LGTM

sayan-biswas (approved):

prabhapa:
qe-approved

sayan-biswas:
@prabhapa I don't think that syntax works for labels also.

/label px-approved
/label docs-approved
/label qe-approved

PR #414, opened 2024-12-06
OCPBUILD-174: Clean Up Containerfiles

adambkaplan:
Remove previous "dev"/rhel7 Dockerfiles, and replace them with an equivalent `Containerfile` based on ubi9. The `Containerfile` uses the suffix `.ubi` as opposed to a specific UBI/RHEL version number to support future upgrades to ubi10 and beyond. The Containerfile uses UBI images from `registry.access.redhat.com`, which does not require a pull secret. This makes it easy for any contributor to pick up and build.

Created a separate Containerfile for OCP builds (`.ocp`), copying the current contents of `Dockerfile.rhel8`. This `Containerfile` uses images from the OpenShift CI system, which are only available to Red Hat employees. The `.rhel8` suffix was removed because it is a source of confusion for new contributors; OCP is currently based on RHEL9 images, and will switch to ubi10 at some point in the future.

PR #425, opened 2025-01-07
OCPBUGS-43812:  k8s bump to 1.30.1

Moebasim:
**Upgraded Kubernetes dependency from 1.28.2 to 1.30.1**

This update upgrades the Kubernetes dependency from 1.28.2 to 1.30.1 to close the version gap that appeared when OpenShift continued with Kubernetes v1.28.2 through version 4.16, even though Kubernetes had moved ahead. OpenShift 4.15 was the last version where OpenShift and Kubernetes were closely aligned. With OpenShift 4.18, Golang was also upgraded from v1.21 to v1.22 to ensure compatibility with the newer Kubernetes release.


Dockerfile
```diff
@@ -32,4 +32,4 @@ RUN ln -s /usr/bin/openshift-builder /usr/bin/openshift-sti-build && \
 LABEL io.k8s.display-name="OpenShift Builder" \
       io.k8s.description="This is a component of OpenShift and is responsible for executing image builds." \
       io.openshift.tags="openshift,builder"
-ENTRYPOINT [ "/usr/bin/entrypoint.sh" ]
+ENTRYPOINT [ "/usr/bin/entrypoint.sh" ]
```
adambkaplan, line 35: Drop this - I don't think it is needed, and it will create a merge conflict with #414

go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/openshift/builder
 
-go 1.21
+go 1.22.0
+
+toolchain go1.22.9
```
adambkaplan, line 5: This will require an update to [ci-operator.yaml](https://github.com/openshift/builder/blob/master/.ci-operator.yaml#L4) - it needs a go 1.22-based builder image (at minimum).

Please also check with the ART team (#forum-ocp-art) that it is OK to add `toolchain` directives to the `go.mod` file.

adambkaplan (changes requested):

PR #446, opened 2025-01-31
OCPBUGS-42893,OCPBUGS-42914: buildah and source-to-image dependency bump to fix - Buildah allows arbitrary directory mount

kunalmemane:
buildah package is bumped from 1.35.4 to 1.35.5 to fix the [CVE-2024-9675](https://github.com/advisories/GHSA-586p-749j-fhwp)

- Buildah bump to v1.35.5
- source-to-image bump to v1.4.0 that resolves indirect docker dependency of buildah and source-to-image.
- storage.DefaultStorageOptions no longer accepts args.
- buildah now requires netavark as an OS dependency.


kunalmemane:
Closed because of Invalid version mapping.

PR #469, opened 2025-04-03
pkg/build/builder/util/dockerfile.Write: reproduce heredocs

nalind:
When writing out the contents of a parsed Dockerfile, attempt to reproduce the heredocs that followed instructions.  Originally encountered by @jlebon while working on https://github.com/openshift/os/pull/1780.


nalind:
Heredocs were first recognized in buildah 1.33, so I expect that this is applicable to release-4.14 and all later branches.

nalind:
But you said!

/remove-lifecycle stale

PR #534, opened 2026-05-12
OCPBUGS-85436: Replace deprecated rand.Seed with pattern-based test assertions

kunalmemane:
## Summary
  - Removed deprecated `rand.Seed(0)` calls from `TestRandomBuildTag`, `TestRandomBuildTagNoDupes`, and `TestContainerName`
  - Replaced exact string assertions with regex pattern-based assertions that validate output format, SHA1 hashing for long names, length limits, and uniqueness
  - Dropped `math/rand` import in favor of `regexp`
  
  ## Why
  `rand.Seed()` is deprecated since Go 1.20. The old tests relied on seeding the global PRNG to get deterministic output, then asserting exact hex values. This was fragile and tested PRNG internals rather than actual function behavior.
  
  ## Test plan
  - [ ] Verify `TestRandomBuildTag` passes — checks short name prefix and format, long name SHA1 hashing and 255-char limit
  - [ ] Verify `TestRandomBuildTagNoDupes` passes — 100 calls produce unique tags
  - [ ] Verify `TestContainerName` passes — output matches expected format pattern

## Summary by CodeRabbit

* **Tests**
  * Enhanced test coverage for build-tag and container-name formatting validation, including prefix verification, hex-suffix format checks, and length constraint verification.

### DIFF
--- a/pkg/build/builder/common_test.go
+++ b/pkg/build/builder/common_test.go
@@ -3,10 +3,10 @@ package builder
 import (
 	"fmt"
 	"io/ioutil"
-	"math/rand"
 	"os"
 	"path/filepath"
 	"reflect"
+	"regexp"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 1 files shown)


kunalmemane:
@coderabbitai review

